PR #5, opened 2024-07-30
fix: repo url in workflow

jonathan-mayer:
## Motivation

fixes the workflow from failing because of a old image repo url

## TODO


JTaeuber (approved):
maybe it works \_(ツ)_/

PR #12, opened 2024-08-06
Feat/basic up and downscaling

jonathan-mayer:
## Motivation

This PR is the basic implementation and structure of the gokubedownscaler

## Changes

Added basic up and downscaling

## Tests done

- Unit tests passed
- tested in staging

## TODO


jonathan-mayer:
For the build and push stage to not fail, the repository for the go-kube-donscaler has to be created on the mtr

JTaeuber:
In order to make this more overseeable we should put all the go code into a seperate folder since we also intend to include a helm chart and other files at some point

jonathan-mayer:
> In order to make this more overseeable we should put all the go code into a seperate folder since we also intend to include a helm chart and other files at some point

As this isn't best practice with go code i would go a slightly different route here. I will make a commit implementing the package structure i would choose.

jonathan-mayer:
I would try to keep this structure.

```
.
├── .gitignore
├── .golangci.yaml
├── .pre-commit-config.yaml
├── CODEOWNERS
├── Dockerfile
├── LICENSE
├── README.md
├── cmd
│   └── kubedownscaler
│       ├── main.go
│       └── main_test.go
├── deployments
│   └── helm
│       └── helm-files-go-here
├── go.mod
├── go.sum
└── internal
    ├── api
    │   └── kubernetes
    │       ├── client.go
    │       └── util.go
    ├── scalable
    │   ├── deployments.go
    │   └── scalableResource.go
    └── values
        ├── duration.go
        ├── layer.go
        ├── layer_test.go
        ├── stringlist.go
        ├── timespan.go
        ├── timespan_test.go
        ├── triStateBool.go
        └── util.go
```

JTaeuber (changes requested):

api/kubernetes/client.go
```diff
@@ -0,0 +1,169 @@
...
+	workload.SetAnnotations(annotations)
+	return nil
+}
+
+// getOriginalReplicas gets the original replicas annotation on the workload. nil is undefined
+func (c client) getOriginalReplicas(workload scalable.Workload) (int, error) {
+	annotations := workload.GetAnnotations()
+	originalReplicasString, ok := annotations[annotationOriginalReplicas]
+	if !ok {
+		return values.Undefined, nil
```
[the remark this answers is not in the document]
jonathan-mayer: This is intentional, because it being unset isn't an "error" as it is a valid state of original replicas. There is a if check for it being "Undefined" where its used that also just "skips" it because of this.

JTaeuber (changes requested):

internal/pkg/values/timespan.go
```diff
@@ -0,0 +1,186 @@
...
+
+	parts := strings.Split(timespanString, " ")
+	weekdaySpan := strings.Split(parts[0], "-")
+	timeSpan := strings.Split(parts[1], "-")
+	timezone := parts[2]
+
+	var err error
+	timespan.timezone, err = time.LoadLocation(timezone)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse timezon: %w", err)
```
JTaeuber, line 89: e missing for timezone. Also I'm not sure if parse is the right word. I think we could leave this as is otherwise.

internal/pkg/values/timespan.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+	timespan.timeFrom, err = time.ParseInLocation("15:04", timeSpan[0], timespan.timezone)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse time: %w", err)
```
JTaeuber, line 93: We should probably include that this error stems from the timeFrom not parsing.

internal/pkg/values/timespan.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+	timespan.timeTo, err = time.ParseInLocation("15:04", timeSpan[1], timespan.timezone)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse time: %w", err)
```
JTaeuber, line 97: We should probably include that this error stems from the timeTo not parsing.

internal/pkg/values/timespan.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+	timespan.weekdayFrom, err = getWeekday(weekdaySpan[0])
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse weekday: %w", err)
```
JTaeuber, line 101: We should probably include that this error stems from the weekdayFrom not parsing.

internal/pkg/values/timespan.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+	timespan.weekdayTo, err = getWeekday(weekdaySpan[1])
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse weekday: %w", err)
```
JTaeuber, line 105: We should probably include that this error stems from the weekdayTo not parsing.

internal/pkg/values/stringlist.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+// StringList is a alias for []string with a Set funciton for the flag package
+type StringList []string
+
+func (s *StringList) Set(value string) error {
```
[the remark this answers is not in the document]
jonathan-mayer: StringList implements the flag.Value interface which has this method signature for the Set function.

internal/pkg/values/timespan.go
jonathan-mayer: I would say that parse is the correct word here, because it is parsing the Timezone form a string.

JTaeuber (approved):
LGTM

PR #17, opened 2024-08-08
feat: multithreaded scanning

### LINKED ISSUE #11: multithreaded scanning
## Issue

Currently the downscaler is scanning each workload one after the other, this will take long when it is scaling hundreds of workloads.

### Problem to solve

Implement multithreading in the scanning process

### Proposal

Run the workload scans each in their own thread. You can use wait groups to wait until all scans have completed.
You might need to change the code to be able to run concurrently.
Test for race conditions!!

### Who can address the issue

Go dev

jonathan-mayer:
## Motivation

Closes #11 

I couldn't test it on a big sample size but it seemed to work and i can't think of anything that wouldn't work concurrently and cause issues.

## Tests done

- [X] tested in staging

## TODO

### DIFF
--- a/cmd/kubedownscaler/main.go
+++ b/cmd/kubedownscaler/main.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"log/slog"
 	"os"
+	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


JTaeuber (approved):
LGTM

PR #19, opened 2024-08-09
feat: added error events on workload

### LINKED ISSUE #16: errors in workload events
## Issue

Currently in a cluster wide installation on a shared cluster users don't know why their configuration is failing without asking for the logs from an administrator.

### Problem to solve

Implement a way to show the user the configuration errors

### Further details

You might need to create a new function in the scalable.Workload interface to have a workload specific implementation

### Proposal

Send the configuration errors / errors with the scaling of the workload as events in the workload.
Try to avoid sending error messages from other modules while keeping it detailed enough to let the user know what is wrong, and what has to be changed.
Keep handling the error on downscaler side, there, keep the detail esp. from errors from other modules.

### Who can address the issue

Go Dev

jonathan-mayer:
## Motivation

Closes #16 

I have decided to just give the user the full error output, even from other modules, since its all just parsing errors and are formatted good enough for now.

## Changes

image

## Tests done

- [X] tested in staging

## TODO

### DIFF
--- a/cmd/kubedownscaler/main.go
+++ b/cmd/kubedownscaler/main.go
@@ -124,12 +124,12 @@ func scanWorkload(workload scalable.Workload, client kubernetes.Client, ctx cont
 		slog.Error("failed to get namespace annotations", "error", err, "workload", workload.GetName(), "namespace", workload.GetNamespace())
 		return false
 	}
-	layerWorkload, err := values.GetLayerFromAnnotations(workload.GetAnnotations(), resourceLogger.Error, ctx)
+	layerWorkload, err := values.GetLayerFromAnnotations(workload.GetAnnotations(), resourceLogger, ctx)
 	if err != nil {
 		slog.Error("failed to parse workload layer from annotations", "error", err, "workload", workload.GetName(), "namespace", workload.GetNamespace())
 		return false
 	}
-	layerNamespace, err := values.GetLayerFromAnnotations(namespaceAnnotations, resourceLogger.Error, ctx)
+	layerNamespace, err := values.GetLayerFromAnnotations(namespaceAnnotations, resourceLogger, ctx)
 	if err != nil {
 		slog.Error("failed to parse namespace layer from annotations", "error", err, "workload", workload.GetName(), "namespace", workload.GetNamespace())
 		return false
--- a/cmd/kubedownscaler/main.go
+++ b/cmd/kubedownscaler/main.go
@@ -139,7 +139,7 @@ func scanWorkload(workload scalable.Workload, client kubernetes.Client, ctx cont
 
 	if layers.GetExcluded() {
 		slog.Debug("workload is excluded, skipping", "workload", workload.GetName(), "namespace", workload.GetNamespace())
-		return false
+		return true
 	}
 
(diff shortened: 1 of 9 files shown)


JTaeuber:
I'll take a look through this tomorrow. I don't think I'll get to it today, sorry.
:(

cmd/kubedownscaler/main.go
```diff
@@ -151,7 +153,7 @@ func scanWorkload(workload scalable.Workload, client kubernetes.Client, ctx cont
 	}
 	if scaling == values.ScalingIgnore {
 		slog.Debug("scaling is ignored, skipping", "workload", workload.GetName(), "namespace", workload.GetNamespace())
-		return false
+		return true
```
JTaeuber, line 156: I did not notice this the first time... Oh well ¯\_(ツ)_/¯

internal/pkg/values/triStateBool.go
```diff
@@ -28,3 +28,6 @@ func (t *triStateBool) String() string {
 	}
 	return fmt.Sprint(t.value)
 }
+
+// IsBoolFlag makes the argument not have a value
+func (b *triStateBool) IsBoolFlag() bool { return true }
```
JTaeuber, line 33: What do you do with this? I don't see it being used anywhere.

internal/pkg/values/util.go
```diff
@@ -26,68 +27,83 @@ const (
...
 		result.ExcludeUntil, err = time.Parse(time.RFC3339, excludeUntil)
 		if err != nil {
+			logEvent.ErrorInvalidAnnotation(annotationExcludeUntil, fmt.Sprintf("failed to parse %q annotaion: %s", annotationExcludeUntil, err.Error()), ctx)
 			return result, fmt.Errorf("failed to parse %q annotation: %w", annotationExcludeUntil, err)
 		}
 	}
 	if forceUptime, ok := annotations[annotationForceUptime]; ok {
 		result.ForceUptime, err = strconv.ParseBool(forceUptime)
 		if err != nil {
+			logEvent.ErrorInvalidAnnotation(annotationForceUptime, fmt.Sprintf("failed to parse %q annotaion: %s", annotationDowntime, err.Error()), ctx)
```
JTaeuber, line 85: Needs to be 
```go
logEvent.ErrorInvalidAnnotation(annotationForceUptime, fmt.Sprintf("failed to parse %q annotaion: %s", annotationForceUptime, err.Error()), ctx).
```
Probably a copy paste error.

JTaeuber (changes requested):

internal/pkg/values/triStateBool.go
jonathan-mayer: This makes the triStateBool type implement the flag.boolFlag interface which allows the cli arguments to have no value eg.: "--explicit-include" instead of "--explicit-include=true"

jonathan-mayer: I did notice that the comment on the function didn't explain this very well, so i changed that. Its function should now be clearer

JTaeuber (approved):
LGTM

Issue #14, opened 2024-08-07
Create Helm Chart for GoKubeDownscaler


JTaeuber:
## Issue

Currently there are no means of deploying the GoKubeDownscaler. We should fix this by creating a helm chart for it.

### Proposal
Create a helm chart for the GoKubeDownscaler with all the necessary components.

### Who can address the issue
People with experience in using helm.

PR #22, opened 2024-08-09
Feat/add helm chart

JTaeuber:
## Motivation
Closes https://github.com/caas-team/GoKubeDownscaler/issues/14.

## Changes
Added new directory and necessary files for the helm chart.

## Tests done
I deployed the chart and it worked so far.
```bash
helm install go-kube-downscaler ./deployments/chart/ -n kubedownscaler-test
NAME: go-kube-downscaler
LAST DEPLOYED: Fri Aug  9 14:34:08 2024
NAMESPACE: kubedownscaler-test
STATUS: deployed
REVISION: 1
TEST SUITE: None
```

However the deployment couldn't pull the image from the mtr yet because our repos are still private.

## TODO


jon4skl (approved):
"LGTM" as you always say :)

JTaeuber:
Hey, so I've refactored a few things. Especially the README has changed completely. The rest was more or less just a bit of cleanup. @jon4skl , @jonathan-mayer have a look and tell me if you think that's alright. If you think it's alright and approve it we can merge it and close the issue for now. There still is a little bit to refactor but I will do that once we have written the main README and provided more downscaling options, so that's alright.

deployments/chart/.helmignore
```diff
@@ -0,0 +1,23 @@
...
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
```
jonathan-mayer, line 23: Do we really need all of these in here, I think this should be cut down a bit

deployments/chart/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v2
+name: go-kube-downscaler
+description: A Helm chart for deploying the go-kube-downscaler
+
+type: application
+version: 0.1.0
+appVersion: "1.0.0"
```
jonathan-mayer, line 7: I think we should choose to use either parentheses everywhere or nowhere

jonathan-mayer:
I think we should add the Workflow for building the chart in this PR

jonathan-mayer:
We might need to evaluate the use of multiple replicas of the downscaler, it should be fine in most cases but if the resource gets updated after the others "got" it from kubernetes but before they updated it it will lead to an error, which will be displayed in the workload events, as well as in the log of the downscaler.

deployments/chart/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+- <a id="tolerations"></a>`tolerations`:
+
+- <a id="affinity"></a>`affinity`: Defines rules for node affinity and pod affinity/anti-affinity. (More info [here](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/))
+
+- <a id="configMapName"></a>`configMapName`: The name of the configmap for the go-kube-downscaler.
+
+- <a id="excludedNamespaces"></a>`excludedNamespaces`: A list of namespaces that are supposed to be excluded from the downscaling process of the go-kube-downscaler.
+
+- <a id="extraConfig"></a>`extraConfig`: Adds additional configurations to the ConfigMap.
```
jonathan-mayer, line 82: This needs to be more precise, mention that it sets the env vars

jonathan-mayer:
Is there a reason why the arguments aren't configurable? the parameter is "missing" in the readme, the values.yaml and the deployment.yaml

jonathan-mayer:
There is already a way to constrict the resources the downscaler is trying to get, i think we should reflect that in the helm chart like we do with the constricted namespaces, add a constrictedResources or alike and only give the service account permission to the set resources

deployments/chart/templates/clusterrole.yaml
```diff
@@ -0,0 +1,166 @@
...
+  name: {{ include "go-kube-downscaler.fullname" . }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ include "go-kube-downscaler.fullname" . }}
+subjects:
+  - kind: ServiceAccount
+    name: {{ include "go-kube-downscaler.serviceAccountName" . }}
+    namespace: {{ .Release.Namespace }}
+{{- end }}
```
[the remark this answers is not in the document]
JTaeuber: This could be done, however this would create two unnecessary if checks which are already covered in the role files.

deployments/chart/templates/clusterrole.yaml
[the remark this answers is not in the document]
JTaeuber: The problem i have with pre-commit atm is that it fails and i don't know why. 
```bash
$ pre-commit run --all-files
Check Yaml...............................................................Failed
- hook id: check-yaml
- exit code: 1

while parsing a flow node
did not find expected node content
  in "deployments/chart/templates/clusterrole.yaml", line 1, column 3
while parsing a flow node
did not find expected node content
  in "deployments/chart/templates/deployment.yaml", line 6, column 7
while constructing a mapping
  in "deployments/chart/templates/serviceaccount.yaml", line 4, column 9
found unhashable key
  in "deployments/chart/templates/serviceaccount.yaml", line 4, column 10
while scanning a simple key
  in "deployments/chart/templates/configmap.yaml", line 8, column 3
could not find expected ':'
  in "deployments/chart/templates/configmap.yaml", line 9, column 3
while parsing a flow node
did not find expected node content
  in "deployments/chart/templates/role.yaml", line 1, column 3
```

JTaeuber:
> Is there a reason why the arguments aren't configurable? the parameter is "missing" in the readme, the values.yaml and the deployment.yaml

I initially omitted the parameter because we don't really have any arguments to pass the go-kube-downscaler. However it needs to go into the deployment.yaml because it carries the constrained namespaces.

I could also add it to the values and readme, but you tell me if you want that.

JTaeuber:
> There is already a way to constrict the resources the downscaler is trying to get, i think we should reflect that in the helm chart like we do with the constricted namespaces, add a constrictedResources or alike and only give the service account permission to the set resources

This is doable. I think i could write a helper for this and make everything even more overseeable. I would name it something like includedResources though because it sounds better. I'll look into it.

deployments/chart/templates/clusterrole.yaml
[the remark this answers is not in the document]
jonathan-mayer: 08afd7f fixes this behaviour

deployments/chart/templates/clusterrole.yaml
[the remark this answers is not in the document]
jonathan-mayer: I would still go ahead with this. An additional 'if-else' statement isn’t such a big deal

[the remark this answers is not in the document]
JTaeuber: I would rather prefer to leave them as is because it makes the roles and rolebindings consistent. If we split the rolebinding from the role creation we have to iterate over all the namespaces again seperately which is not too pretty.

deployments/chart/templates/deployment.yaml
```diff
@@ -18,6 +26,13 @@ spec:
       containers:
         - name: {{ .Chart.Name }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          {{- if $hasArgs }}
+          args:
+          {{- toYaml .Values.arguments | nindent 10 }}
+          {{- if .Values.constrainedDownscaler }}
+          - --namespace={{ join "," .Values.constrainedNamespaces }}
+          {{- end }}
```
jonathan-mayer, line 34: I would add something like this for includedResources (`--include-resources`)

deployments/chart/values.yaml
```diff
@@ -4,6 +4,21 @@ image:
...
+  - deployments
+  - statefulsets
+  - daemonsets
+  - rollouts
+  - horizontalpodautoscalers
+  - jobs
+  - cronjobs
+  - scaledobjects
+  - stacks
+  - poddisruptionbudgets
```
jonathan-mayer, line 20: The default is just deployments, defaults should be in the values.yaml as comments.

``` suggestion
includedResources:
  - deployments
#  - statefulsets
#  - daemonsets
#  - rollouts
#  - horizontalpodautoscalers
#  - jobs
#  - cronjobs
#  - scaledobjects
#  - stacks
#  - poddisruptionbudgets
```

OR

``` suggestion
includedResources:
  - deployments
```

jonathan-mayer (changes requested):
once these are resolved the PR should be ready

deployments/chart/values.yaml
jonathan-mayer: * actually don't just use the suggestion, it doesn't seem to have realised i was selecting multiple lines

deployments/chart/README.md
```diff
@@ -57,7 +57,7 @@ Here is a brief overview of all the values contained in this chart:
 
 - <a id="arguments"></a>`arguments`: Defines command line arguments for the container.
 
-- <a id="includedResources"></a>`includedResources`: Defines all the resources that the go-kube-downscaler is supposed to be able to handle.
+- <a id="includedResources"></a>`includedResources`: Defines all the resources that the go-kube-downscaler is supposed to be able to handle. This list can not be empty.
```
jonathan-mayer, line 60: Actually it can, the downscaler just won't do anything. (unless something in the helm chart makes it not work)
The default set in the main.go should be overwritten with an empty string list, so it will just loop through nothing every scan

jonathan-mayer (approved):
LGTM 👍

jonathan-mayer:
After the merge, we should make the first release

JTaeuber:
Sure but let's wait for @jon4skl to have a quick look through the changes as well. We don't have any pressure regarding time.

jon4skl (approved):

jonathan-mayer:
I will merge this, but i will actually wait with the first release until #31 and probably also #28 is merged

PR #30, opened 2024-08-25
Feat/Add Resources (StatefulSets, DaemonSets, HPA, PodDisruptionBudgets, CronJobs, Jobs, Keda ScaledObjects)

samuel-esp:
## Motivation

Added the following resources, already present in the original py-kube-downscaler

- StatefulSets
- DaemonSets
- HPA
- PodDisruptionBudgets
- CronJobs
- ScaledObjects

## Changes

- Workload Interface now becomes a father interface for many sub-interfaces
- Refactored logic inside DownscaleWorkload and UpscaleWorkload, the type of scaling will be different based on the "Family" of the resource and it will be automatically understood thanks to the child interface
- Added dynamic client interface to handle future crds (Keda, Argo, ...)
- 

## Tests done

Built image and tested on a mock cluster. Will add also unit tests asap

## TODO


samuel-esp:
I did have problems with keda regarding these 2 dependencies when building the image

k8s.io/apimachinery v0.30.3 -> 0.31 solves the issue

sigs.k8s.io/controller-runtime v0.17.5 -> 0.19 solves the issue

i kept the mod file as similar as possible to the original. I dont know if the release is target for 1.30 (even though 1.31 should be retrocompatible)

PR #31, opened 2024-08-25
Feat/Add Resources (StatefulSets, DaemonSets, HPA, PodDisruptionBudgets, CronJobs, Jobs, Keda ScaledObjects)

samuel-esp:
## Motivation

Added the following resources, already present in the original py-kube-downscaler

- StatefulSets
- DaemonSets
- HPA
- PodDisruptionBudgets
- CronJobs
- ScaledObjects

PS: I closed the old PR, and re opened here a clean one with squashed commits and the commit message written using the convention. This commit was made with lint disabled, I'll use it when more commits made with the new lints are merged inside the main (it was preventing me to push due to errors in modules I didn't want to touch for this task)

## Changes

- Workload Interface now becomes a father interface for many sub-interfaces
- Refactored logic inside DownscaleWorkload and UpscaleWorkload, the type of scaling will be different based on the "Family" of the resource and it will be automatically understood thanks to the child interface
- Added dynamic client interface to handle future crds (Keda, Argo, ...)
- 

## Tests done

Built image and tested on a mock cluster. Will add also unit tests asap

## TODO

I kept the mod file as similar as possible to the original. I dont know if the k8s release you are targeting for is the 1.30 (most probably the answer is yes seeing the starting dependencies, even though 1.31 apis should be retrocompatible).

I did have problems with keda regarding these 2 dependencies when building the image. So we should check how to solve these compatibilities problem

- k8s.io/apimachinery v0.30.3 -> 0.31beta solves the issue
- sigs.k8s.io/controller-runtime v0.17.5 -> 0.19 solves the issue but requieres k8s.io dependencies to be at 0.31. Probabily using different golang base image different than alpine will solve the problems at 0.18.5

these problems arised because Keda releaeses under 2.15.1 are incompatible with the 0.30.3 apis. Upgrading to 0.31 and controller-runtime 0.19 solved everything and I was able to correctly test it 


internal/pkg/scalable/scaledobjects.go
```diff
@@ -0,0 +1,98 @@
...
+	if s.Annotations == nil {
+		s.Annotations = map[string]string{}
+		s.Annotations[kedaPausedReplicasAnnotation] = stringReplicas
+	} else {
+		s.Annotations[kedaPausedReplicasAnnotation] = stringReplicas
+	}
+}
+
+// RemovePauseScaledObjectAnnotation will remove the keda pause annotation
+func (s ScaledObject) RemovePauseScaledObjectAnnotation() error {
```
jonathan-mayer, line 57: This func is pretty useless, and has a lot of unneeded overhead.
The functionality could be achived just with: (we don't need an error if it doesn't exist, we just care that is no longer exists after calling this)
``` go
// RemovePauseScaledObjectAnnotation will remove the keda pause annotation
func (s ScaledObject) RemovePauseScaledObjectAnnotation() {
    delete(s.Annotations, kedaPausedReplicasAnnotation)
} 
```

Possibly everything to do with this should be refactored to not have to neeed this funciton at all

internal/pkg/scalable/scaledobjects.go
```diff
@@ -0,0 +1,98 @@
...
+	if s.Annotations == nil {
+		return fmt.Errorf("no annotation inside this scaledobject")
+	} else {
+		if _, ok := s.Annotations[kedaPausedReplicasAnnotation]; !ok {
+			fmt.Errorf("there is no keda scaling object pause annotation to remove")
```
jonathan-mayer, line 62: missing return

internal/pkg/scalable/scaledobjects.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// GetPauseScaledObjectAnnotationReplicasIfExistsAndValid gets the value of keda pause annotations. It returns the int value and true if the annotations exists and it is well formatted, otherwise it returns a fake value and false
+func (s ScaledObject) GetPauseScaledObjectAnnotationReplicasIfExistsAndValid() (int, bool, error) {
+	if s.Annotations == nil {
+		return 1, false, nil
+	}
+	if pausedReplicasStr, ok := s.Annotations[kedaPausedReplicasAnnotation]; !ok {
+		return 1, false, nil
+	} else {
```
jonathan-mayer, line 76: all of these else blocks are not needed. these could all be replaced like this:
``` go
if something {
    // do something
    return anotherThing
}
// everything within else block goes here
```
try to avoid using else at all.
try to always use guard clauses.

jonathan-mayer (changes requested):
I think I will leave the review at this for now. It may doesn't seem much but the requested changes should be applied anywhere with similar problems. Also i think this should be refactored. With the changes most of the funcs do almost nothing. 
Install pre commit, if it gives an error it is because of an error that should be resolved. If you want to know which error it is you can use `pre-commit run --all-files`, it will give an output of what problems exist. Also run `go test ./...` sometimes or just let pre commit do it automatically on commits.

jonathan-mayer:
You also don't have to make an docker image, you can just run the downscaler locally with `go run ./cmd/kubedownscaler -k=/path/to/kubeconfig` (the kubeconfig has to have the permissions and the context set to the one you want to use)

internal/pkg/scalable/scaledobjects.go
samuel-esp: I didn't understand this point, do you prefer to call directly the map delete from the upscaling function without referring to this method inside the ScaledObject class?

internal/api/kubernetes/client.go
```diff
@@ -88,6 +95,79 @@ func (c client) GetWorkloads(namespaces []string, resourceTypes []string, ctx co
 
 // DownscaleWorkload downscales the workload to the specified replicas
 func (c client) DownscaleWorkload(replicas int, workload scalable.Workload, ctx context.Context) error {
+	switch w := workload.(type) {
+	case scalable.BatchWorkload:
+		return c.DownscaleBatchWorkload(w, ctx)
```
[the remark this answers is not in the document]
samuel-esp: That's ok, I'll make both functions and scaled classes private (like right now we have for deployments)

internal/pkg/scalable/scaledobjects.go
samuel-esp: Refactored as you said, in this way. Tell me if it is ok

```
func (s ScaledObject) GetPauseScaledObjectAnnotationReplicasIfExistsAndValid() (int, bool, error) {
	if s.Annotations == nil {
		return 1, false, nil
	}
	if pausedReplicasStr, ok := s.Annotations[kedaPausedReplicasAnnotation]; ok {
		pausedReplicas, err := strconv.Atoi(pausedReplicasStr)
		if err != nil {
			return 1, false, fmt.Errorf("invalid value for annotation %s: %w", kedaPausedReplicasAnnotation, err)
		}
		return pausedReplicas, true, nil
	}

	return 1, false, nil
}
```

internal/pkg/scalable/scaledobjects.go
jonathan-mayer: well, the function would only be the delete, so i would refactor the upscaling func to allow for this behavior without needing this one-liner func.

internal/pkg/scalable/scalableResource.go
```diff
@@ -3,23 +3,33 @@ package scalable
...
+	"statefulsets":             getStatefulSets,
+	"cronjobs":                 getCronJobs,
+	"jobs":                     getJobs,
+	"daemonsets":               getDaemonSets,
+	"poddisruptionbudgets":     getPodDisruptionBudgets,
+	"horizontalpodautoscalers": getHorizontalPodAutoscalers,
+	"scaledobjects":            getScaledObjects,
 }
 
 // Workload is a interface for a scalable resource. It holds all resource specific functions
```
jonathan-mayer, line 35:
```suggestion
// Workload is a interface for a scalable resource. It holds all shared resource specific functions
```

internal/api/kubernetes/client.go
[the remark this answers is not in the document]
samuel-esp: completed

internal/pkg/scalable/scaledobjects.go
samuel-esp: functions deleted, used the default already implemented inside Workload Interface

internal/pkg/scalable/scaledobjects.go
samuel-esp: refactored

jonathan-mayer:
I know this is a pretty big change, but could you refactor it like this:
```
// Workload is a interface for a scalable resource. It holds all resource specific functions
type Workload interface {
	// GetAnnotations gets the annotations of the resource
	GetAnnotations() map[string]string
	// GetNamespace gets the namespace of the resource
	GetNamespace() string
	// GetName gets the name of the resource
	GetName() string
	// GetUID gets the uid of the workload
	GetUID() types.UID
	// GetObjectKind gets the ObjectKind of the workload
	GetObjectKind() schema.ObjectKind
	// SetAnnotations sets the annotations on the resource. Changes won't be made on kubernetes until update() is called
	SetAnnotations(annotations map[string]string)
	// Update updates the resource with all changes made to it. It should only be called once on a resource
	Update(clientset *kubernetes.Clientset, dynamicClient dynamic.Interface, ctx context.Context) error
	// TODO comment
	ScaleUp()  error // error only for parsing or configuration problems
	// TODO comment
	ScaleDown(downscaleReplicas) error // error only for parsing or configuration problems
	// GetCurrentReplicas gets the current amount of replicas of the resource
	GetCurrentReplicas() (int, error)
}
```

also remove the funcs in workload that then nolonger need to be accessible.

Upscale and downscale basically do:
``` go
// DownscaleAppWorkload is the dedicated downscale function for apps/v1 workloads, excluding daemonsets
func (c client) DownscaleWorkload(replicas int, workload scalable.Workload, ctx context.Context) error {
	workload.ScaleDown()
	err = workload.Update(c.clientset, c.dynamicClient, ctx)
	if err != nil {
		return fmt.Errorf("failed to update workload: %w", err)
	}
	slog.Debug("successfully scaled down workload", "workload", workload.GetName(), "namespace", workload.GetNamespace())
	return nil
}
```

we then no longer need the other interfaces for every type, the client.RemoveOriginalReplicas, Set and Get can be private funcs within the util.go file in the scalable package

cronjobs just set the suspend
the rest just sets the replicas in whatever way it needs to eg. 
deployments sets replicas, 
hpa additionally checks for if replicas < 1
keda sets the annotation(i think)

i know this is a pretty big change but it will make this waaaaayy easier to read and understand. the scaledown dowscale replicas can be expicitly ignored with a _ in the func that dont need it like the cronjob

samuel-esp:
Of course, just give me some time and it will be ready!

internal/api/kubernetes/client.go
```diff
@@ -88,87 +92,35 @@ func (c client) GetWorkloads(namespaces []string, resourceTypes []string, ctx co
...
-func (c client) removeOriginalReplicas(workload scalable.Workload) {
-	annotations := workload.GetAnnotations()
-	delete(annotations, annotationOriginalReplicas)
-	workload.SetAnnotations(annotations)
-}
-
 // AddErrorEvent creates or updates a new event on the workload
 func (c client) AddErrorEvent(reason, id, message string, workload scalable.Workload, ctx context.Context) error {
-	hash := sha1.Sum([]byte(fmt.Sprintf("%s.%s", id, message)))
+	hash := sha256.Sum256([]byte(fmt.Sprintf("%s.%s", id, message)))
```
jonathan-mayer, line 123: oh yea i forgot about that, it was fixed in other PRs but they aren't merged yet because the other code owners were on vacation

jonathan-mayer (commented):
Ill have a look over it first thing tomorrow

internal/api/kubernetes/client.go
samuel-esp: That was the reason why I disabled the lint. I didn't want to modify this class before it was merged on main

internal/pkg/scalable/cronjobs.go
```diff
@@ -0,0 +1,87 @@
+package scalable
+
+import (
+	"context"
+	"fmt"
```
jonathan-mayer, line 5: if the suspend is nil the cronjob is not suspended right? so why then care that it doesn't have the attribute. i would just do it like this:
```suggestion
// ScaleUp upscale the resource when the downscale period ends
func (c cronJob) ScaleUp() error {
	const suspend = false

	currentSuspendIsTrue, err := c.getSuspend()
	if err != nil {
		return fmt.Errorf("failed to get current replicas for workload: %w", err)
	}
	if !currentSuspendIsTrue {
		slog.Debug("workload is already upscaled, skipping", "workload", c.GetName(), "namespace", c.GetNamespace())
		return nil
	}

	c.setSuspend(suspend)
	return nil
}
```

internal/pkg/scalable/deployments.go
```diff
@@ -27,23 +33,75 @@ type deployment struct {
...
+func (d deployment) getCurrentReplicas() (int, error) {
 	replicas := d.Spec.Replicas
 	if replicas == nil {
 		return 0, errNoReplicasSpecified
 	}
 	return int(*d.Spec.Replicas), nil
 }
 
+// ScaleUp upscale the resource when the downscale period ends
+func (d deployment) ScaleUp() error {
```
jonathan-mayer, line 58: again use *deployment everywhere where values are directly modified

internal/pkg/scalable/deployments.go
```diff
@@ -27,23 +33,75 @@ type deployment struct {
 	*appsv1.Deployment
 }
 
-// SetReplicas sets the amount of replicas on the resource. Changes won't be made on kubernetes until update() is called
-func (d deployment) SetReplicas(replicas int) {
+// setReplicas sets the amount of replicas on the resource. Changes won't be made on kubernetes until update() is called
+func (d deployment) setReplicas(replicas int) error {
+	if replicas > math.MaxInt32 || replicas < math.MinInt32 {
```
jonathan-mayer, line 38: I'd probably check for if the replicas is < 0 thought, as that is the lowest valid value. If you do so also change the error message below. Also new errors (errors which don't wrap another error) should be defined as a var at the top of the file (we will change those to custom errors later).

internal/pkg/scalable/deployments.go
```diff
@@ -27,23 +33,75 @@ type deployment struct {
...
+		return fmt.Errorf("failed to get original replicas for workload: %w", err)
+	}
+	if originalReplicas == values.Undefined {
+		slog.Debug("original replicas is not set, skipping", "workload", d.GetName(), "namespace", d.GetNamespace())
+		return nil
+	}
+	if originalReplicas == currentReplicas {
+		slog.Debug("workload is already at original replicas, skipping", "workload", d.GetName(), "namespace", d.GetNamespace())
+		return nil
+	}
```
jonathan-mayer, line 74: I'd actually remove all the things used only for debug in all of the scaling funcs. 

```suggestion
	originalReplicas, err := getOriginalReplicas(d)
	if err != nil {
		return fmt.Errorf("failed to get original replicas for workload: %w", err)
	}
	if originalReplicas == values.Undefined {
		slog.Debug("original replicas is not set, skipping", "workload", d.GetName(), "namespace", d.GetNamespace())
		return nil
	}
```

jonathan-mayer (changes requested):
Ok, I think if you should change these for all resources before I review further. I would wait with writing the tests until i say that the code is fine.

internal/pkg/scalable/daemonsets.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get daemonsets: %w", err)
+	}
+	for _, item := range daemonsets.Items {
+		results = append(results, daemonSet{&item})
+	}
+	return results, nil
+}
+
+// daemonSet is a wrapper for batch/v1.cronJob to implement the scalableResource interface
```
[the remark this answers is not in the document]
samuel-esp: completed

jonathan-mayer (changes requested):
still have to review:
scalableResources.go
scaledobjects
statefulsets
util.go

jonathan-mayer:
I dont know if you already did, but could you retest these changes on a cluster. (you can also wait until the review is finished)

jonathan-mayer (commented):
Some small fixes, and i pushed a refactor/fix of the scaledObjects. I will still look into a way of not having that dynamic client and replacing it with a normal client with a custom scheme. You could already start writing tests for the ScaleUp and ScaleDown functions, as those shouldn't change any of the other aspects of the downscaling. It would also be awsome if you could test these resources manually in a cluster.

samuel-esp:
I'll try to test everything asap both in the real cluster and test classes, I might be a little bit busier in the next days but I'll do my best

internal/pkg/scalable/scaledobjects.go
```diff
@@ -0,0 +1,115 @@
...
+		so := &kedav1alpha1.ScaledObject{}
+		if err := runtime.DefaultUnstructuredConverter.FromUnstructured(item.Object, so); err != nil {
+			return nil, fmt.Errorf("failed to convert unstructured to scaledobject: %w", err)
+		}
+		results = append(results, &scaledObject{so})
+	}
+	return results, nil
+}
+
+// scaledObject is a wrapper for keda.sh/v1alpha1.horizontalPodAutoscaler to implement the Workload interface
```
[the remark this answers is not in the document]
samuel-esp: yeah! wrong copy paste for comment

internal/pkg/scalable/scalableResource.go
```diff
@@ -36,10 +47,10 @@ type Workload interface {
 	GetObjectKind() schema.ObjectKind
 	// SetAnnotations sets the annotations on the resource. Changes won't be made on kubernetes until update() is called
 	SetAnnotations(annotations map[string]string)
```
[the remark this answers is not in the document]
samuel-esp: @jonathan-mayer I thought this method was inherited from a k8s api interface, isn't it? When I tried his twin method (GetAnnotations()) in the real cluster, it worked out of the box even without actual implementations within the codebase.  If I set it to private it will require me to do a custom implementations on every resource

samuel-esp:
@jonathan-mayer I will try to start writing tests between this evening or tomorrow, look at the unresolved topic, I have a small question on a method

internal/pkg/scalable/scalableResource.go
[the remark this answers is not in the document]
jonathan-mayer: Oh your right, my bad.

internal/pkg/scalable/statefulsets.go
```diff
@@ -73,7 +73,7 @@ func (s *statefulSet) ScaleUp() error {
 	return nil
 }
 
-// ScaleDown downscale the resource when the downscale period starts
+// ScaleDown downscale the resource
```
jonathan-mayer, line 74: i would change resource to workload  and it would either have to be
`// ScaleDown downscales the workload` or `// ScaleDown scales down the workload`

internal/pkg/scalable/scalableResource.go
```diff
@@ -36,10 +47,10 @@ type Workload interface {
...
-	// SetReplicas sets the amount of replicas on the resource. Changes won't be made on kubernetes until update() is called
-	SetReplicas(replicas int)
-	// GetCurrentReplicas gets the current amount of replicas of the resource
-	GetCurrentReplicas() (int, error)
 	// Update updates the resource with all changes made to it. It should only be called once on a resource
-	Update(clientset *kubernetes.Clientset, ctx context.Context) error
+	Update(clientset *kubernetes.Clientset, dynamicClient dynamic.Interface, ctx context.Context) error
+	// ScaleUp upscale the resource when the downscale period ends
```
jonathan-mayer, line 52: the comments for ScaleUp and ScaleDown should also be updated in the workload interface

jonathan-mayer (changes requested):

samuel-esp:
Added unit tests, unfortunately didn't have time to test in a real cluster yet

jonathan-mayer:
LGTM I will merge after I refactored the tests a bit to have some more test cases.

internal/pkg/scalable/deployments.go
```diff
@@ -3,48 +3,89 @@ package scalable
...
+	err = d.setReplicas(*originalReplicas)
+	if err != nil {
+		return fmt.Errorf("failed to set original replicas for workload: %w", err)
+	}
+	removeOriginalReplicas(d)
+	return nil
+}
+
+// ScaleDown downscale the resource
+func (d *deployment) ScaleDown(downscaleReplicas int) error {
```
jonathan-mayer, line 72: I just realised we need to change something back here, currently the ScaleDown with original replicas would set the orginal replicas to the downscaled replicas in the second scan being downscaled. For all resources that use orignial replicas, we need to add back the check if CurrentReplicas == DownscaleReplicas

jonathan-mayer (dismissed):
LGTM, We should let @JTaeuber have a look over it aswell, just to make sure i didn't miss anything

internal/api/kubernetes/client.go
```diff
@@ -88,89 +93,37 @@ func (c client) GetWorkloads(namespaces []string, resourceTypes []string, ctx co
 
 // DownscaleWorkload downscales the workload to the specified replicas
 func (c client) DownscaleWorkload(replicas int, workload scalable.Workload, ctx context.Context) error {
-	originalReplicas, err := workload.GetCurrentReplicas()
+	err := workload.ScaleDown(replicas)
 	if err != nil {
-		return fmt.Errorf("failed to get original replicas for workload: %w", err)
+		return fmt.Errorf("failed to complete the scale down process on workload: %w", err)
```
JTaeuber, line 98:
```suggestion
		return fmt.Errorf("failed to scale down the workload: %w", err)
```

internal/api/kubernetes/client.go
```diff
@@ -88,89 +93,37 @@ func (c client) GetWorkloads(namespaces []string, resourceTypes []string, ctx co
...
-		return nil
-	}
-
-	workload.SetReplicas(replicas)
-	c.setOriginalReplicas(originalReplicas, workload)
-	err = workload.Update(c.clientset, ctx)
+	err = workload.Update(c.clientsets, ctx)
 	if err != nil {
-		return fmt.Errorf("failed to update workload: %w", err)
+		return fmt.Errorf("failed to update the workload after completing the scaling down process: %w", err)
```
JTaeuber, line 102:
```suggestion
		return fmt.Errorf("failed to update the workload: %w", err)
```

internal/api/kubernetes/client.go
```diff
@@ -88,89 +93,37 @@ func (c client) GetWorkloads(namespaces []string, resourceTypes []string, ctx co
...
 	return nil
 }
 
 // UpscaleWorkload upscales the workload to the original replicas
 func (c client) UpscaleWorkload(workload scalable.Workload, ctx context.Context) error {
-	currentReplicas, err := workload.GetCurrentReplicas()
+	err := workload.ScaleUp()
 	if err != nil {
-		return fmt.Errorf("failed to get current replicas for workload: %w", err)
+		return fmt.Errorf("failed to complete the scale up process on workload: %w", err)
```
JTaeuber, line 112:
```suggestion
		return fmt.Errorf("failed to scale up the workload: %w", err)
```

internal/api/kubernetes/client.go
```diff
@@ -88,89 +93,37 @@ func (c client) GetWorkloads(namespaces []string, resourceTypes []string, ctx co
...
-		slog.Debug("workload is already at original replicas, skipping", "workload", workload.GetName(), "namespace", workload.GetNamespace())
-		return nil
-	}
-
-	workload.SetReplicas(originalReplicas)
-	c.removeOriginalReplicas(workload)
-	err = workload.Update(c.clientset, ctx)
-	if err != nil {
-		return fmt.Errorf("failed to update workload: %w", err)
+		return fmt.Errorf("failed to update the workload after completing the scaling up process: %w", err)
```
JTaeuber, line 116:
```suggestion
		return fmt.Errorf("failed to update the workload: %w", err)
```

JTaeuber (changes requested):

JTaeuber (changes requested):

internal/pkg/scalable/daemonsets.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	return results, nil
+}
+
+// daemonSet is a wrapper for batch/v1.cronJob to implement the Workload interface
+type daemonSet struct {
+	*appsv1.DaemonSet
+}
+
+// ScaleUp upscale the resource
```
JTaeuber, line 33:
```suggestion
// ScaleUp scales the resource up
```

internal/pkg/scalable/daemonsets.go
```diff
@@ -0,0 +1,55 @@
...
+func (d *daemonSet) ScaleUp() error {
+	delete(d.Spec.Template.Spec.NodeSelector, labelMatchNone)
+	return nil
+}
+
+// ScaleDown downscale the resource
```
JTaeuber, line 39:
```suggestion
// ScaleDown scales the resource down
```

JTaeuber (changes requested):

internal/pkg/scalable/deployments.go
```diff
@@ -3,48 +3,93 @@ package scalable
...
+// getCurrentReplicas gets the current amount of replicas of the resource
+func (d *deployment) getCurrentReplicas() (int, error) {
 	replicas := d.Spec.Replicas
 	if replicas == nil {
 		return 0, errNoReplicasSpecified
 	}
 	return int(*d.Spec.Replicas), nil
 }
 
+// ScaleUp upscale the resource
```
JTaeuber:
```suggestion
// ScaleUp scales the resource up
```

JTaeuber (changes requested):

internal/pkg/scalable/horizontalpodautoscalers.go
```diff
@@ -0,0 +1,97 @@
...
+	"context"
+	"fmt"
+	"log/slog"
+	"math"
+
+	appsv1 "k8s.io/api/autoscaling/v2"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// getHorizontalPodAutoscalers is the getResourceFunc for horizontalPodAutoscaler
```
JTaeuber, line 13:
```suggestion
// getHorizontalPodAutoscalers is the getResourceFunc for horizontalPodAutoscalers
```

internal/pkg/scalable/horizontalpodautoscalers.go
```diff
@@ -0,0 +1,97 @@
...
+func getHorizontalPodAutoscalers(namespace string, clientsets *Clientsets, ctx context.Context) ([]Workload, error) {
+	var results []Workload
+	poddisruptionbudgets, err := clientsets.Kubernetes.AutoscalingV2().HorizontalPodAutoscalers(namespace).List(ctx, metav1.ListOptions{TimeoutSeconds: &timeout})
```
JTaeuber, line 16:
```suggestion
	horizontalpodautoscalers, err := clientsets.Kubernetes.AutoscalingV2().HorizontalPodAutoscalers(namespace).List(ctx, metav1.ListOptions{TimeoutSeconds: &timeout})
```

internal/pkg/scalable/horizontalpodautoscalers.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get horizontalpodautoscalers: %w", err)
+	}
+	for _, item := range poddisruptionbudgets.Items {
```
JTaeuber, line 20:
```suggestion
	for _, item := range horizontalpodautoscalers.Items {
```

JTaeuber (changes requested):

internal/pkg/scalable/jobs.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	batch "k8s.io/api/batch/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// getDeployments is the getResourceFunc for Deployments
```
JTaeuber, line 11:
```suggestion
// getJobs is the getResourceFunc for Jobs
```

JTaeuber (changes requested):

internal/pkg/scalable/poddisruptionbudgets.go
```diff
@@ -0,0 +1,149 @@
...
+	maxUnavailable := p.getMaxUnavailableInt()
+	minAvailable := p.getMinAvailableInt()
+	if maxUnavailable != values.Undefined {
+		if maxUnavailable == downscaleReplicas {
+			slog.Debug("workload is already scaled down, skipping", "workload", p.GetName(), "namespace", p.GetNamespace())
+			return nil
+		}
+		err := p.setMaxUnavailable(downscaleReplicas)
+		if err != nil {
+			return fmt.Errorf("failed to set minAvailable for workload: %w", err)
```
JTaeuber, line 121:
```suggestion
			return fmt.Errorf("failed to set maxUnavailable for workload: %w", err)
```

JTaeuber (changes requested):

internal/pkg/scalable/scalableResource.go
```diff
@@ -4,25 +4,36 @@ import (
...
+	errNoMinReplicasSpecified          = errors.New("error: workload has no minimum replicas set")
+	errBoundOnScalingTargetValue       = errors.New("error: replicas can only be set to a 32-bit integer >= 0 (or >= 1 for HPAs)")
 )
 
-// getResourceFunc is a function that gets a specific resource as a scalableResource
-type getResourceFunc func(namespace string, clientset *kubernetes.Clientset, ctx context.Context) ([]Workload, error)
+// getResourceFunc is a function that gets a specific resource as a Workload
+type getResourceFunc func(namespace string, clientsets *Clientsets, ctx context.Context) ([]Workload, error)
 
 // GetResource maps the resource name to a implementation specific getResourceFunc
```
JTaeuber, line 24:
```suggestion
// GetResource maps the resource name to an implementation specific getResourceFunc
```

JTaeuber (changes requested):

internal/pkg/scalable/scaledobjects.go
```diff
@@ -0,0 +1,101 @@
...
+
+// setPauseAnnotation sets the value of keda pause annotations
+func (s *scaledObject) setPauseAnnotation(value int) {
+	if s.Annotations == nil {
+		s.Annotations = map[string]string{}
+	}
+	s.Annotations[annotationKedaPausedReplicas] = strconv.Itoa(value)
+}
+
+// ScaleUp scales the resource up when the downscale period ends
```
JTaeuber, line 58:
```suggestion
// ScaleUp scales the resource up
```

internal/pkg/scalable/scaledobjects.go
```diff
@@ -0,0 +1,101 @@
...
+		delete(s.Annotations, annotationKedaPausedReplicas)
+		removeOriginalReplicas(s)
+		return nil
+	}
+	s.setPauseAnnotation(*originalReplicas)
+	removeOriginalReplicas(s)
+	return nil
+}
+
+// ScaleDown scales down the workload
```
JTaeuber, line 78:
```suggestion
// ScaleDown scales the workload down
```
is not uniform

JTaeuber (changes requested):

JTaeuber (approved):
LGTM

PR #33, opened 2024-09-03
fix: fixed timezone america to north america

jon4skl:
## Motivation

changed "America" to "North America" to make sure using correct continents

## TODO


jonathan-mayer:
It is actually just America, see: https://data.iana.org/time-zones/tzdb/zone1970.tab

PR #35, opened 2024-09-05
Feat/overlapping timespan check

jon4skl:
## Motivation

Since there are relative and absolute timezones we need to make sure that they dont overlap in any circumstances.

## Tests done

Tests based on my changes are written and they are working

## TODO

### DIFF
--- a/internal/pkg/values/timespan.go
+++ b/internal/pkg/values/timespan.go
@@ -169,8 +169,8 @@ func (t relativeTimeSpan) inLocation(timezone *time.Location) []relativeTimeSpan
 			timezone:    timezone,
 			timeFrom:    sameDays.timeFrom.Add(24 * time.Hour),
 			timeTo:      sameDays.timeTo.Add(24 * time.Hour),
-			weekdayFrom: sameDays.weekdayFrom - 1,
-			weekdayTo:   sameDays.weekdayTo - 1,
+			weekdayFrom: getWeekdayBefore(sameDays.weekdayFrom),
+			weekdayTo:   getWeekdayBefore(sameDays.weekdayTo),
 		}
 		result = append(result, daysBefore)
 	}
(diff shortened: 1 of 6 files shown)


internal/pkg/values/timespan.go
```diff
@@ -143,14 +145,58 @@ func (t relativeTimeSpan) isTimeOfDayInRange(timeOfDay time.Time) bool {
...
+			weekdayFrom: sameDays.weekdayFrom + 1,
+			weekdayTo:   sameDays.weekdayTo + 1,
+		}
+		result = append(result, daysAfter)
+	}
+	return result
+}
+
+// Check if timeFrom skipped to the previous day
+func isTimeFromSkippedToPreviousDay(timeFrom time.Time) bool {
```
jonathan-mayer, line 191: i would change this to `hasTimeFromSkippedToPreviousDay` and maybe put it in util.go.
Also the comment should be in this format: `// functionName description`

Same with `isTimeToSkippedToNextDay`.

jonathan-mayer (changes requested):

internal/pkg/values/timespan.go
```diff
@@ -143,14 +145,58 @@ func (t relativeTimeSpan) isTimeOfDayInRange(timeOfDay time.Time) bool {
...
+		timeTo:      t.timeTo.In(timezone),
+	}
+	result = append(result, sameDays)
+	if isTimeFromSkippedToPreviousDay(sameDays.timeFrom) {
+		daysBefore := relativeTimeSpan{
+			timezone:    timezone,
+			timeFrom:    sameDays.timeFrom.Add(24 * time.Hour),
+			timeTo:      sameDays.timeTo.Add(24 * time.Hour),
+			weekdayFrom: sameDays.weekdayFrom - 1,
+			weekdayTo:   sameDays.weekdayTo - 1,
```
jonathan-mayer, line 173: needs to check if goes under 0 and loop back to 6, same with the +1 below

internal/pkg/values/util.go
```diff
@@ -157,3 +157,23 @@ func isTimeFromSkippedToPreviousDay(timeFrom time.Time) bool {
...
+
+// getWeekdayAfter returns the day after (it automatically checks if value is below 0 or above 6)
+func getWeekdayAfter(weekday time.Weekday) time.Weekday {
+	if weekday == 6 {
+		weekday = 0
+		return weekday
+	} else {
+		return weekday + 1
+	}
+}
```
jonathan-mayer, line 179: If we were to do it like this we would use guard lauses instead of the else and return 0 directly instead of assinging it to weekday and returning weekday. But I would actaully go a different route here using modulo. (`return (weekday+1)%7` and `return (weekday+7-1)%7`)

jonathan-mayer:
LGTM, once you resolve the merge conflicts and @JTaeuber had a look over it its ready to merge

Issue #41, opened 2024-09-10
Send message in PR when merging would create release


jonathan-mayer:
## Issue

Currently if someone isn't careful when reviewing a PR the workflows could automatically create an unwanted Release.

### Problem to solve

We need a way of easily being warned when merging a PR would result in a new Release

### Further details

Depends on #42 

### Proposal

Create a Workflow or adjust the check_version workflow to check if merging the PR would result in a new release. 

It's important that even if the change was made after the PR opened (by committing a Chart.yaml version bump after opening the PR) the message still has to be sent.

It would be even better if the maintainer had to accept these risks before being able to merge. eg. a manually triggered pipeline that has to complete before merging is allowed, or by commenting or reacting with an emoji to the message.

PR #44, opened 2024-09-11
Enhance CI Workflow: Check for Version Bump and appVersion Changes

Fovty:
## Motivation

This change improves the existing CI workflow by adding a check for both version bumps and changes to the `appVersion` in the `Chart.yaml`. It aims to ensure that any changes to key versioning information are caught and flagged during the pull request process, providing maintainers the ability to block merges if necessary.

- Closes https://github.com/caas-team/GoKubeDownscaler/issues/41

## Changes

- Added logic to check for differences in `appVersion` between the PR and the `main` branch.
- The PR will be blocked if an `appVersion` change is detected.
- Added a condition to skip the workflow if the `release` label is applied.
- The workflow triggers on pull request events (opened, synchronize, labeled).

## Tests done

- Verified that the workflow properly detects version changes in `Chart.yaml`.
- Tested the condition to skip the workflow based on the `release` label.
- Tested the Git diff functionality to detect changes in `appVersion` field.
- Ensured the workflow triggers correctly on PR events like opened, synchronize, and labeled.

## TODO

- Ensure that the branch protection rules for the main branch enforce the `Version Bump Check` status check.
- The workflow requires **write access** to contents and pull requests to set statuses and post comments. Make sure the following permissions are enabled for GitHub Actions:
  - **Contents**: `write`
  - **Pull Requests**: `write`


.github/workflows/check-version-bump.yml
jonathan-mayer, line 1: I think we should keep the workflow file extensions the same, so ".yaml" instead of ".yml". Also we could rename it to something like "check-for-release.yaml"

.github/workflows/check-version-bump.yml
```diff
@@ -0,0 +1,46 @@
+name: Check for Version Bump and appVersion in Chart.yaml
```
jonathan-mayer, line 1: I would also rename this to "Check for new release"

.github/workflows/check-version-bump.yml
```diff
@@ -0,0 +1,46 @@
...
+
+on:
+  pull_request:
+    types: [opened, synchronize, labeled]
+
+jobs:
+  check_version_bump:
+    name: Version Bump Check
```
jonathan-mayer, line 9: Maybe also this to something like "Check for relase"

jonathan-mayer (approved):
Just some naming, maybe @JTaeuber will find something but the rest looks good to me

jonathan-mayer (approved):
LGTM, lets have @JTaeuber have a look over it

JTaeuber (approved):
I will change the rest on my branch. Other than that it looks good.

PR #45, opened 2024-09-12
feat: add pre commit workflow

### LINKED ISSUE #43: Check pre commit workflow
## Issue

Many Contributors might not have pre commit installed or forgot to execute `pre-commit install` in the repository. This results in problems if the reviewers forget to check if pre commit was run properly and doesnt give an error, which results in branches from the commit in main all having the pre commit problems, having to resolve them in each branch which then again leads to more merge conflicts.

### Further details

The [pre-commit/action](https://github.com/pre-commit/action) could be used, although it is in maintenance only mode. You could also possibly find a maintained action that does this.
Or check out pre-commit ci [for organization repositories](https://pre-commit.ci/)

### Proposal

We need an automatic way of running pre commit and seeing if there are any problems. This could be resolved using a Workflow, a GitHub app or anything that works here. The check doesn't need to, but it would be nice if it would automatically push changes made to files the checks. 

### Other links/references

https://pre-commit.ci/

jonathan-mayer:
## Motivation

Closes #43 

## Changes

Adds a workflow that automatically runs pre-commit. It fails if pre commit fails and pushes changes if any were made

## Tests done

- [X] Tested in fork

## TODO


jonathan-mayer:
One thing to consider is that pre commit also automatically runs all tests, so we kind of don't need the go test workflow, what do you think about this @Fovty @JTaeuber @jon4skl

.github/workflows/pre_commit.yaml
```diff
@@ -0,0 +1,45 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup Go environment
+        uses: actions/setup-go@v5.0.2
+        with:
+          go-version-file: "go.mod"
+      - name: Install dependencies
+        run: |
+          # install go modules
```
JTaeuber, line 21:
```suggestion
          # install required go packages
```

.github/workflows/pre_commit.yaml
```diff
@@ -0,0 +1,45 @@
...
+          go install mvdan.cc/gofumpt@latest
+          # install golangci-lint
+          curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin
+      - name: Run pre-commit
+        uses: pre-commit/action@v3.0.1
+        env:
+          SKIP: "no-commit-to-branch" # if not skipped, will always fail on main
+      - name: Push changes
+        run: |
+          git config --global user.name "github-actions[bot]"
```
JTaeuber, line 33:
```suggestion
          git config --global user.name "github-actions"
```

.github/workflows/pre_commit.yaml
```diff
@@ -0,0 +1,45 @@
...
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
```
JTaeuber, line 34:
```suggestion
          git config --global user.email "github-actions@github.com"
```

JTaeuber (changes requested):

JTaeuber (approved):
LGTM

PR #54, opened 2024-09-16
feat: let workflow trigger on label removal

JTaeuber:
## Motivation
The check_for_release workflow didn't trigger on label deletion.

## Changes
Added unlabeled keyword.


jonathan-mayer (approved):
I hope this works

PR #70, opened 2024-11-18
Chore/change prettier hook

JTaeuber:
## Motivation
Change outdated pre-commit hooks and auto update them.

## Changes
* Combined gitignores
* Changed prettier hook
* updated pre-commit hooks
* Fixed readme link


jonathan-mayer (commented):
Interesting that pip is already installed.
I would normally have not wanted to have it installed twice but it shouldn't since pip is just going to say that the requirement is already met.

One consideration would be to move the autoupdate under the check so we don't have to install it manually.
The pre commit workfow would automatically run on the updated versions of the hooks once the workflow pushes the pre-commit-config.yaml changes.

jonathan-mayer:
While we are here we could also consider changing the check for release workflow to only run when the pull request is trying to merge to main.

jonathan-mayer (approved):
LFineTM

PR #73, opened 2024-11-30
Fix: Add missing URL scheme in README link

Fovty:
## Motivation

Fix the README link to ensure it is clickable and redirects correctly to the `py-kube-downscaler` GitHub repository.

## Changes

- Added the missing `https://` scheme to the README link.

## Tests done

- Verified the link is clickable and redirects to the correct repository.


jonathan-mayer (approved):
This is already fixed in feat/add-docusaurus, but since it's still going to take a while to merge that we can go ahead and merge this to fix it faster. Thanks @Fovty :+1:!

Issue #68, opened 2024-11-05
Allow for synchronous operation


jonathan-mayer:
## Issue

Currently the downscaler doesn't support multiple replicas/running multiple instances on the same workloads.

### Problem to solve

If a workload is modified by another instance after the instance already got the workload from Kubernetes the downscaler will throw an error because the resourceVersion changed.
The same errors occur if the workload was modified manually or by something else during scaling.

### Proposal

I don't think it is possible to handle both multiple instances and manual interference perfectly, at least not without erroring or logging sometimes or in some kind of way.
Because of this we should really change the error handling for this error to give more information to the user (possibly link to documentation on the topic).
The error could also be changed to a info or its logging could be turned off by a cli argument as it isn't broken and will just retry in the next scan cycle.

If we want the downscaler to allow for multiple instances we need to implement some way to make this error less likely/not occur with multiple instances of the downscaler.
One way to implement this would be using Kubernetes' leader election using leases. This would let only one downscaler would scale at a time. This would be redundant and stop the error from occurring because of multiple downscaler instances entirely.

Combining both of these would result in the error only being thrown on outside/manual intervention making it much less likely to occur.
Additionally when it occurs the user doesn't need to search for help with this intermittent error because it explains exactly what happend.

### Who can address the issue

Go dev with some Kubernetes know-how

### Other links/references

[./docs/troubleshooting.md#synchronous-operation](https://github.com/caas-team/GoKubeDownscaler/blob/main/docs/troubleshooting.md#synchronous-operation)
https://github.com/kubernetes/client-go/blob/master/examples/leader-election/main.go#L112
https://kubernetes.io/docs/concepts/architecture/leases/#custom-workload

jonathan-mayer:
There should be an option to run in "standalone mode". Or the non-"standalone mode" should be only activated when an option is passed

samuel-esp:
I've just found a similar problem on Py-Kube-Downscaler. Basically the problem you are describing could happen not only if you are using more than 1 replicas, but also when the downscaler iteration is too long. I still need to check if the code we have here could have the same problem, I think the concurrency already implemented helps in mitigating this

caas-team/py-kube-downscaler#111

jonathan-mayer:
> The problem could be present since the resources is retrieved before being updated
> 
> https://github.com/caas-team/GoKubeDownscaler/blob/main/cmd/kubedownscaler/main.go#L113
> 
[...]

PR #74, opened 2024-12-14
Feat/retries on conflict error

samuel-esp:
## Motivation

Give the user the possibility to choose how to handle HTTP 409 conflict errors. Such conflicts typically occur when another entity (such as an HPA, CI/CD pipeline, or manual intervention) modifies a resource just before KubeDownscaler processes it

See #68 or https://github.com/caas-team/py-kube-downscaler/issues/111

## Changes

- Introduced `--max-retries-on-conflict` argument like in Py-Kube-Downscaler
- Introduced GetWorkload() function to handle the use case when the downscaler needs to retrieve a single Kubernetes resource (before it was only possible to get a list of resources, i.e `kubectl get deploy -n default`). the old GetWorkload() was renamed to GetWorkloads() to reflect the changes
- Introduced a new function GetResourceType() that returns the resource type (string)
- Refactored the main loop to be able to use `--max-retries-on-conflict`

## Tests done

- Unit Tests

## TODO

- [x] Refactored docs
- [ ] Added more unit tests on this specific use case


jonathan-mayer (commented):
Also just so you know, the workflows were broken for forks (thats why they were failing). We've fixed it but it will now run every workflow twice in this and the other pr. If you want you can rebase the branches with main and the errors will go away.

internal/pkg/scalable/cronjobs.go
```diff
@@ -21,11 +21,27 @@ func getCronJobs(namespace string, clientsets *Clientsets, ctx context.Context)
...
+// getCronJob is the getResourceFunc for a single CronJob
+func getCronJob(name string, namespace string, clientsets *Clientsets, ctx context.Context) (Workload, error) {
+	var result Workload
+	cronjob, err := clientsets.Kubernetes.BatchV1().CronJobs(namespace).Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		return result, fmt.Errorf("failed to get cronjob: %w", err)
+	}
+	result = &suspendScaledWorkload{&cronJob{cronjob}}
+	return result, nil
+}
```
jonathan-mayer, line 33: I really think we shouldn't have a whole other function just to get just 1 of the resource.

```suggestion
// getCronJobs is the getResourceFunc for CronJobs
func getCronJobs(name, namespace string, clientsets *Clientsets, ctx context.Context) ([]Workload, error) {
    var results []Workload
    if name != "" {
        cronjob, err := clientsets.Kubernetes.BatchV1().CronJobs(namespace).Get(ctx, name, metav1.GetOptions{})
        if err != nil {
            return result, fmt.Errorf("failed to get cronjob: %w", err)
        }
        results = append(results, &suspendScaledWorkload{&cronJob{cronjob}})
        return results
    }

    cronjobs, err := clientsets.Kubernetes.BatchV1().CronJobs(namespace).List(ctx, metav1.ListOptions{TimeoutSeconds: &timeout})
    if err != nil {
        return nil, fmt.Errorf("failed to get cronjobs: %w", err)
    }
    for _, item := range cronjobs.Items {
        results = append(results, &suspendScaledWorkload{&cronJob{&item}})
    }
    return results, nil
}
```

I think we should keep the client.GetWorkload and GetWorkloads in single functions. This will also help abstract this configuration away.

cmd/kubedownscaler/main.go
```diff
@@ -121,18 +125,40 @@ func main() {
...
-
-				err := scanWorkload(workload, client, ctx, layerCli, layerEnv)
-				if err != nil {
-					slog.Error("failed to scan workload", "error", err, "workload", workload.GetName(), "namespace", workload.GetNamespace())
-					return
+				attempts := 0
+				for {
+					err := scanWorkload(workload, client, ctx, layerCli, layerEnv)
+					if err != nil {
+						if strings.Contains(err.Error(), "the object has been modified") {
```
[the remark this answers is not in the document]
samuel-esp: That's elegant jonathan, will refactor to this. Great find!

samuel-esp:
@jonathan-mayer rebased and included the previous suggestions. Just a couple of things to note:

1. Pre-Commit failed because the specific `getResourceFunc` for each resource has duplicate code" detected (which is true because the logic is shared across them). So, eventually, we should try to think about a strategy to tackle that
2. Pre-Commit also suggested to lower a bit `startScanning` complexity. What would you refactor inside that?

jonathan-mayer:
Ill have a look over it. 

> 1. Pre-Commit failed because the specific getResourceFunc for each resource has duplicate code" detected (which is true because the logic is shared across them). So, eventually, we should try to think about a strategy to tackle that

I do get that it detects it as duplicate, but im not entirely sure how to avoid it. I think in theory it is avoidable by just having list and get in separate functions. Another way would also be to refactor to use the dynamic client again, although i would still like to avoid that.

> 2. Pre-Commit also suggested to lower a bit startScanning complexity. What would you refactor inside that?

I think we could refactor the function called in the go routine out.

samuel-esp:
1. I perfectly agree with you, I don't see anything wrong with having them duplicate. I will go nolint on them
2. I'll extract the go routine from the function

samuel-esp:
@jonathan-mayer I refactored the go func out of the `startScanning` function however I'm not sure what could be the best name for the that

jonathan-mayer:
> @jonathan-mayer I refactored the go func out of the `startScanning` function however I'm not sure what could be the best name for the that. `attemptScan` was the best I came up with

i think the current function name should be good for now. Although i would not put a anonymous function in the attemptScan function, but instead just run the attemptScan function in a goroutine so `go attemptScan()`. Also just a heads up, it might take some time until i get around to review this.

samuel-esp:
Don't worry Jonathan, absolutely no rush take your time

jonathan-mayer:
I think we should change up the GetResource funcs again. They should only handle listing and then every scalableResouce should have a re-get function which regets itself from kubernetes. it could like something like this:

```
func (c *cronJob) Reget(clientsets *Clientsets, ctx context.Context) {
	*c, err = clientsets.Kubernetes.BatchV1().CronJobs(c.Namespace).Get(c.Name, ctx,, metav1.GetOptions{})
	if err != nil {
		// TODO handle error
	}
}
```

jonathan-mayer (changes requested):

internal/pkg/util/config.go
```diff
@@ -102,6 +104,12 @@ func (c *RuntimeConfiguration) ParseConfigFlags() {
 		"",
 		"the annotation to use instead of creation time for grace period (optional)",
 	)
+	flag.IntVar(
+		&c.MaxRetriesOnConflict,
+		"max-retries-on-error",
```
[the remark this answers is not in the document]
samuel-esp: nice catch!

samuel-esp:
> I think we should change up the GetResource funcs again. They should only handle listing and then every scalableResouce should have a re-get function which regets itself from kubernetes. it could like something like this:
> 
> ```go
> func (c *cronJob) Reget(clientsets *Clientsets, ctx context.Context) {
> 	*c, err = clientsets.Kubernetes.BatchV1().CronJobs(c.Namespace).Get(c.Name, ctx,, metav1.GetOptions{})
> 	if err != nil {
> 		// TODO handle error
> 	}
> }
> ```

Implemented, the linter still complain about 2 things:
- deployment/statefulset class duplication
- RegetWorkload and its specific methods returns an interface type

I mean the first one could be true, the second one it seems legit to return an interface type in that case

jonathan-mayer (changes requested):
looked at main and clientgo

internal/api/kubernetes/client.go
```diff
@@ -130,6 +132,18 @@ func (c client) GetWorkloads(name string, namespaces, resourceTypes []string, ct
 	return results, nil
 }
 
+// RegetWorkload gets the workload again to ensure the latest state.
+func (c client) RegetWorkload(name, namespace, resourceType string, ctx context.Context) (scalable.Workload, error) {
```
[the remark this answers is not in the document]
samuel-esp: fine, I'll refactor this

samuel-esp:
refactored the way you suggested, still having "duplicate" suggestion on linting. I think we can't do much about that

cmd/kubedownscaler/main.go
```diff
@@ -215,6 +205,65 @@ func startScanning(
...
+	}
+
+	if !scanSucceded && config.MaxRetriesOnConflict > 0 {
+		slog.Warn("max retries reached, will try again in the next scan", "workload",
+			workload.GetName(),
+			"namespace",
+			workload.GetNamespace())
+
+		return
+	}
```
jonathan-mayer, line 255: im not too much of a fan of putting logs in if statements and this message wouldn't fit into the case where `once` is specified (the main proposed use case of the retries argument). so i would try to change up these messages so that it is a bit more generic. 

`slog.Error(failed to scan workload, "tries/attempts", config.MaxRetriesOnConflict+1, ...)`

\*i would choose "tries" or "attempts", not both


we should be able to put this after the for loop. then instead of having the `scanSucceeded` bool we can just return where the `scanSucceeded = true` is now

internal/pkg/scalable/cronjobs.go
```diff
@@ -29,6 +28,18 @@ type cronJob struct {
...
+// Reget regets the resource from the Kubernetes API.
+func (c *cronJob) Reget(clientsets *Clientsets, ctx context.Context) error {
+	singleCronJob, err := clientsets.Kubernetes.BatchV1().CronJobs(c.Namespace).Get(ctx, c.Name, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to get cronjob: %w", err)
+	}
+
+	c.CronJob = singleCronJob
+
+	return nil
```
jonathan-mayer, line 40:
```suggestion
    var err error
	c.CronJob, err = clientsets.Kubernetes.BatchV1().CronJobs(c.Namespace).Get(ctx, c.Name, metav1.GetOptions{})
	if err != nil {
		return fmt.Errorf("failed to get cronjob: %w", err)
	}

	return nil
```

this should be done with every resource

internal/pkg/scalable/workload.go
```diff
@@ -81,6 +81,8 @@ type Workload interface {
 	ScaleUp() error
 	// ScaleDown scales down the workload
 	ScaleDown(downscaleReplicas int32) error
+	// Reget regets the workload to ensure the latest state
+	Reget(clientsets *Clientsets, ctx context.Context) error
```
jonathan-mayer, line 85: i think this should be in the scalableResource interface

jonathan-mayer (changes requested):

cmd/kubedownscaler/main.go
```diff
@@ -215,6 +205,64 @@ func startScanning(
...
+
+	defer waitGroup.Done()
+
+	scanSucceded := false
+
+	for retry := 0; retry <= config.MaxRetriesOnConflict; retry++ {
+		err := scanWorkload(workload, client, ctx, layerCli, layerEnv, config)
+		if err != nil {
+			if !(strings.Contains(err.Error(), registry.OptimisticLockErrorMsg)) {
+				slog.Error("failed to scan workload", "error", err, "workload", workload.GetName(), "namespace", workload.GetNamespace())
```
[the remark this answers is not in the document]
samuel-esp: I got your point, I refactored that accordingly

cmd/kubedownscaler/main.go
```diff
@@ -187,7 +187,13 @@ func startScanning(
...
 			waitGroup.Add(1)
 
-			go attemptScan(client, ctx, layerCli, layerEnv, config, waitGroup.Done, workload)
+			go func() {
+				err = attemptScan(client, ctx, layerCli, layerEnv, config, waitGroup.Done, workload)
+				if err != nil {
+					slog.Error("failed to scan workload", "error", err, "workload", workload.GetName(), "namespace", workload.GetNamespace())
+					return
+				}
+			}()
```
jonathan-mayer, line 196: oh, can you actually redo what you did [here](https://github.com/caas-team/GoKubeDownscaler/pull/74#discussion_r1905199891)

cmd/kubedownscaler/main.go
```diff
@@ -186,17 +188,11 @@ func startScanning(
...
-				slog.Debug("scanning workload", "workload", workload.GetName(), "namespace", workload.GetNamespace())
-
-				defer waitGroup.Done()
-
-				err := scanWorkload(workload, client, ctx, layerCli, layerEnv, config)
+				err = attemptScan(client, ctx, layerCli, layerEnv, config, waitGroup.Done, workload)
 				if err != nil {
 					slog.Error("failed to scan workload", "error", err, "workload", workload.GetName(), "namespace", workload.GetNamespace())
 					return
 				}
```
jonathan-mayer, line 195: from https://github.com/caas-team/GoKubeDownscaler/pull/74#discussion_r1969015079
> or we could wrap calling the attemptScan function in a anonymous func which handles waitGroup.Done and error handling

jonathan-mayer (changes requested):

cmd/kubedownscaler/main.go
samuel-esp: yes!

jonathan-mayer (dismissed):
I think this should be good to go

samuel-esp:
Done

jonathan-mayer (approved):

samuel-esp:
@jonathan-mayer rebased and cleaned history

internal/pkg/scalable/scaledobjects.go
```diff
@@ -14,7 +15,7 @@ const (
 	annotationKedaPausedReplicas = "autoscaling.keda.sh/paused-replicas"
 )
 
-// getScaledObjects is the getResourceFunc for Keda ScaledObjects.
+// getScaledObjects is the getResourceFunc for Keda ScaledObjects. //nolint:dupl.
```
jonathan-mayer, line 18: do we still need this no lint here if we have the whole file nolinted fo rthis linter

```suggestion
// getScaledObjects is the getResourceFunc for Keda ScaledObjects.
```

internal/pkg/scalable/stacks.go
```diff
@@ -9,7 +9,9 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// getStacks is the getResourceFunc for Zalando Stacks.
+// getStacks is the getResourceFunc for Zalando Stacks. //nolint:dupl.
+//
+//nolint:dupl // necessary to handle different workload types separately
```
jonathan-mayer, line 14:
```suggestion
// getStacks is the getResourceFunc for Zalando Stacks.
//nolint:dupl // necessary to handle different workload types separately
```

jonathan-mayer (changes requested):

internal/pkg/scalable/stacks.go
jonathan-mayer: oh acutally, can you remove the //nolint:dupl on that func entirely. the file already has the nolint for this

samuel-esp: done!

samuel-esp:
I think everthing is ready now

jonathan-mayer (dismissed):

jonathan-mayer:
Resolved merge conflicts. Feel free to merge whenever

PR #75, opened 2024-12-18
Feat/leader election mechanism using leases

samuel-esp:
## Motivation

In order to avoid conflicts among multiple replicas when modifying resources (see #68) it is necessary to implement a leader election mechanism that elects a leader among the available replicas. The feature should be implemented using Leases (see [Kubernetes Leases Explained](https://kubernetes.io/docs/concepts/architecture/leases/))

## Changes

- Introduced a new atomic variable inside `main.go` that stores the state of the replica (`isLeader`)  
- Introduced a nested GoRoutine that manages lease lifecycle, the GoRoutine periodically checks the lease status and creates/renews the lease. When a termination signal is sent by the user or by another entity and the replica is currently the leader, it handles a graceful termination by deleting the current lease  
- Implemented 2 new functions in `client.go` to manage lease lifecycle used by the nested GoRoutine mentioned above (`CreateOrUpdateLease` and `DeleteLease`). The first function calls the function `GetCurrentNamespaceFromFile` from `util.go` (which reads the ServiceAccount used by the downscaler) to understand the namespace where it needs to create the lease object  
- Added a new `Role` and `RoleBinding` to the Helm chart that allows the downscaler to create, get, update and delete `Leases` inside its own namespace  

## Tests done

- Live Tests

## TODO

- [x] Fine Tuning Lease Renewal Time
- [x] Unit Testing
- [ ] Further Live Testing


internal/api/kubernetes/client.go
```diff
@@ -207,3 +218,86 @@ func (c client) addWorkloadEvent(eventType, reason, id, message string, workload
 	}
 	return nil
 }
+
+// CreateOrUpdateLease attempts to acquire and maintain a lease for leadership.
```
jonathan-mayer, line 222: The kubernetes package/client should only have very basic functions which can in most cases basically be linked to a single api request to the cluster. So I would make a leader package for all the logic and only have a function here to create/update a lease object.

jonathan-mayer (changes requested):
Had a first look at the changes. What I've found is that there is actually a leader election package from k8s. I think we should use that one since it makes leader election easier and less verbose. 
https://pkg.go.dev/k8s.io/client-go/tools/leaderelection

I've also found this articel which gives an example of how to use it https://dev.to/sklarsa/how-to-add-kubernetes-powered-leader-election-to-your-go-apps-57jh

deployments/chart/templates/_helpers.tpl
```diff
@@ -267,3 +267,19 @@ Create defined permissions for roles
...
+    - coordination.k8s.io
+  resources:
+    - leases
+  verbs:
+    - get
+    - create
+    - watch
+    - list
+    - update
+    - delete
```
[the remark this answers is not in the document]
samuel-esp: Done!

internal/api/kubernetes/client.go
samuel-esp: refactored using the "https://pkg.go.dev/k8s.io/client-go/tools/leaderelection" package

samuel-esp:
> Had a first look at the changes. What I've found is that there is actually a leader election package from k8s. I think we should use that one since it makes leader election easier and less verbose. https://pkg.go.dev/k8s.io/client-go/tools/leaderelection
> 
> I've also found this articel which gives an example of how to use it https://dev.to/sklarsa/how-to-add-kubernetes-powered-leader-election-to-your-go-apps-57jh

I've refactored the PR to use the native "leader-election" package, unit tests are ok, live testing as well but to be sure I would prefer a live testing from another person as well

Also, there is still something to improve in the code structure I think, so I'm waiting for the review

Finally, we should discuss about the "lease-role", I think it will be easier with a separate role

deployments/chart/templates/_helpers.tpl
```diff
@@ -267,3 +267,19 @@ Create defined permissions for roles
...
+{{- end -}}
```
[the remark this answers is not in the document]
samuel-esp: @jonathan-mayer I'll leave this discussion open here

jonathan-mayer:
Resolved the left over merge conflicts (which were basically already solved, no idea why GitHub showed them. Either way ima have a look over this in the next days. Thanks @samuel-esp!

cmd/kubedownscaler/main.go
```diff
@@ -80,15 +89,59 @@ func main() {
...
+	leaderelection.RunOrDie(ctx, leaderelection.LeaderElectionConfig{
+		Lock:            lease,
+		ReleaseOnCancel: true,
+		LeaseDuration:   60 * time.Second,
+		RenewDeadline:   15 * time.Second,
+		RetryPeriod:     5 * time.Second,
+		Callbacks: leaderelection.LeaderCallbacks{
+			OnStartedLeading: func(ctx context.Context) {
+				slog.Info("started leading")
+				run(ctx)
```
jonathan-mayer, line 123: i dont see why we cant just call the loop function directly here

cmd/kubedownscaler/main.go
```diff
@@ -80,15 +89,59 @@ func main() {
...
+				slog.Info("stopped leading")
+			},
+			OnNewLeader: func(identity string) {
+				slog.Info("new leader elected", "identity", identity)
+			},
+		},
+	})
+}
+
+func loop(client kubernetes.Client, ctx context.Context, layerCli, layerEnv *values.Layer, config *util.RuntimeConfiguration) {
```
jonathan-mayer, line 135: would rename to `startScanning` or similar

cmd/kubedownscaler/main.go
```diff
@@ -80,15 +89,59 @@ func main() {
...
+	signal.Notify(ch, os.Interrupt, syscall.SIGTERM)
+
+	go func() {
+		<-ch
+		cancel()
+	}()
+
+	leaderelection.RunOrDie(ctx, leaderelection.LeaderElectionConfig{
+		Lock:            lease,
+		ReleaseOnCancel: true,
```
jonathan-mayer, line 116: > If you set this to true, you must ensure all code guarded by this lease has successfully completed prior to cancelling the context, or you may have two processes simultaneously acting on the critical path.

could this cause problems when a sigterm comes in and a scan is underway? or do all the api calls just get canceled via the context before this can cause issues?

deployments/chart/templates/_helpers.tpl
```diff
@@ -267,3 +267,6 @@ Create defined permissions for roles
 {{- end }}
 {{- end }}
 {{- end }}
+{{/*
+Create defined permissions for lease role
+*/}}
```
jonathan-mayer, line 272:
```suggestion
```

jonathan-mayer (changes requested):
Great work @samuel-esp. One big thing we have to watch out for though is that I think its really important to be able to easily test the downscaler locally, this is currently no longer possible because it wants to get the namespace of the current pod. 

I think what we should do is put the GetNamespace func call directly above the client.CreateLease call. If the GetNamespace func returns empty, log info that it seems not to be running on cluster and it will run without leader election. Additionally put an guard clause above the client.CreateLease call and just run the program directly:
```go
downscalerNamespace, err := kubernetes.GetCurrentNamespaceFromFile()
if err != nil {
  slog.Info("running outside of cluster; skipping leader election", "error", err)
  startScanning()
  return
}

lease, err := client.CreateLease(leaseName, downscalerNamespace)
if err != nil {
  slog.Error("failed to create lease", "error", err)
  os.Exit(1)
}

// ...
```

deployments/chart/templates/_helpers.tpl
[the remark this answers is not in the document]
jonathan-mayer: @JTaeuber

samuel-esp:
> Great work @samuel-esp. One big thing we have to watch out for though is that I think its really important to be able to easily test the downscaler locally, this is currently no longer possible because it wants to get the namespace of the current pod.
> 
> I think what we should do is put the GetNamespace func call directly above the client.CreateLease call. If the GetNamespace func returns empty, log info that it seems not to be running on cluster and it will run without leader election. Additionally put an guard clause above the client.CreateLease call and just run the program directly:
> 
> ```go
> downscalerNamespace, err := kubernetes.GetCurrentNamespaceFromFile()
> if err != nil {
>   slog.Info("running outside of cluster; skipping leader election", "error", err)
>   startScanning()
>   return
> }
> 
> lease, err := client.CreateLease(leaseName, downscalerNamespace)
> if err != nil {
>   slog.Error("failed to create lease", "error", err)
>   os.Exit(1)
> }
> 
> // ...
> ```

@jonathan-mayer thanks a lot for quickly reviewing the PR. I've improved this behavior you mentioned and now if the leader election couldn't be completed the main process won't stop. I added a couple of logs messages to make the user aware of the potential conflicts this situation could lead to 

I refactored the code according to the suggestions, I left a couple of discussions open (for the dedicated lease `Role`, and for the `ReleaseOnCancel` behavior)

cmd/kubedownscaler/main.go
```diff
@@ -89,25 +83,30 @@ func main() {
...
+		return
 	}
 
-	defer cancel()
+	lease, err := client.CreateLease(leaseName, downscalerNamespace)
+	if err != nil {
+		slog.Warn("failed to create lease", "error", err)
+		slog.Warn("proceeding without leader election, this may cause multiple instances to conflict when modifying the same resources")
+		startScanning(client, ctx, &layerCli, &layerEnv, config)
+	}
```
jonathan-mayer, line 103: CreateLease only returns an error when os.Hostname does. We shouldn't continue at all if we seem to be running in cluster but something as simple as the hostname fails.
The Warning about proceeding without leader election should be placed in the guard clause above, and we should just log the error and exit here.

jonathan-mayer (changes requested):
Thanks for resolving the threads so fast, I have to test this still, but i think the code looks good and I dont see why it wouldn't work. We should maybe still discuss the left open threads before merging real quick.

cmd/kubedownscaler/main.go
jonathan-mayer: I will still have to test this either way, but ive looked a bit into this. This should not cause any problems, since all api calls should instantly cancel once the context is. If in the future we see any problems arising from this we can simply set this to false. This would cause the RunOrDie function call to stop blocking when the context is canceled which in turn returns the main, forcibly terminating all other goroutines. After that the lease would expire after the LeasePeriod, giving the program even more time to finish up. Only reason I am against doing this right now is that it does mean that there will be more downtime and it really shouldn't cause any problems like it is right now.

deployments/chart/templates/_helpers.tpl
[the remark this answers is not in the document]
jonathan-mayer: I'll still wait to hear something back from @JTaeuber before resolving this, but I also just realised that not having this separate would have dire consequences in a non-cluster-wide installation where the downscalers own namespace is not selected to be scaled. I currently can't come up with any solutions which would have any benefit over just having this separate.

cmd/kubedownscaler/main.go
```diff
@@ -80,15 +83,64 @@ func main() {
...
+		<-sigs
+		cancel()
+	}()
+
+	leaderelection.RunOrDie(ctx, leaderelection.LeaderElectionConfig{
+		Lock:            lease,
+		ReleaseOnCancel: true,
+		LeaseDuration:   60 * time.Second,
+		RenewDeadline:   15 * time.Second,
+		RetryPeriod:     5 * time.Second,
```
jonathan-mayer, line 118: Also, on the topic of the TODO for fine tuning these values. 
I think it would be smart to something more like this:
```suggestion
		LeaseDuration:   30 * time.Second,
		RenewDeadline:   20 * time.Second,
		RetryPeriod:     5 * time.Second,
```
Its common to have the RenewDeadline be 2/3 of the LeaseDuration. I think the RetryPeriod being 5s is fine, since its not like the downscaler needs to be super high availability and we already send a lot of api requests as it is. I think the Lease Duration being 30s is a good middle ground for the downscaler, it means that the leader still has 3-4 chances at renewing and also 30s being the default Scan interval this means that at least when using the default interval there is basically 0 downtime between scans. I would go against retrieving these values from the specified scan interval, because if these values are too low it will cause a lot more leader switches and in turn unnecessary overhead.

cmd/kubedownscaler/main.go
```diff
@@ -80,15 +83,64 @@ func main() {
...
 		slog.Error("failed to scan over workloads",
 			"error", err,
 			"config", config,
-			"CliLayer", layerCli,
-			"EnvLayer", layerEnv,
+			"CliLayer", *layerCli,
+			"EnvLayer", *layerEnv,
 		)
 		os.Exit(1)
 	}
```
jonathan-mayer, line 146: Instead of exiting here, we should return an error and handle it in main. 

If the func was called from within the LeaderElection we can then cancel the context and not exit, which releases the lease instead of having to wait for the lease to expire.

In the case where we run without leader election, we can simply print the error, cancel the context and exit. No if err check or letting the context stop the program required.

jonathan-mayer (changes requested):
Turns out I did have a bit more to say about this 🙃

cmd/kubedownscaler/main.go
samuel-esp:
```go
lease, err := client.CreateLease(leaseName, downscalerNamespace)
if err != nil {
slog.Warn("failed to create lease", "error", err)
os.Exit(1)
}

defer cancel()
```
@jonathan-mayer to resolve this and exit from the process when the hostname retrieval fails, I had to move the `defer cancel()` below since the linter won't allow me to commit for  `exitAfterDefer: os.Exit will exit, and `defer cancel()` will not run (gocritic)`

jonathan-mayer: i think i wouldn't move the defer cancel, im not sure if the linter picks up on it, if not we could do a nolint for the os.Exit but i think we should just be able to cancel the context before the exit.

```go
lease, err := client.CreateLease(leaseName, downscalerNamespace)
if err != nil {
slog.Warn("failed to create lease", "error", err)
cancel()
os.Exit(1)
}
```

cmd/kubedownscaler/main.go
samuel-esp: refactored

samuel-esp:
@jonathan-mayer I left another discussion open for a linter suggestion that may go against one of the suggestion you said before about initializing context

cmd/kubedownscaler/main.go
samuel-esp: I left this open since it could conflict with what you said there

> @[jonathan-mayer](https://github.com/jonathan-mayer) jonathan-mayer [Jan 26, 2025](https://github.com/caas-team/GoKubeDownscaler/pull/75#discussion_r1929628831)
> 
> put this directly after initializing the context

deployments/chart/templates/_helpers.tpl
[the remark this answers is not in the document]
JTaeuber: Yeah we should keep this seperate for now so we don't allow unnecessary actions.

cmd/kubedownscaler/main.go
```diff
@@ -80,15 +83,64 @@ func main() {
...
+	leaderelection.RunOrDie(ctx, leaderelection.LeaderElectionConfig{
+		Lock:            lease,
+		ReleaseOnCancel: true,
+		LeaseDuration:   30 * time.Second,
+		RenewDeadline:   20 * time.Second,
+		RetryPeriod:     5 * time.Second,
+		Callbacks: leaderelection.LeaderCallbacks{
+			OnStartedLeading: func(ctx context.Context) {
+				slog.Info("started leading")
+				startScanning(client, ctx, &layerCli, &layerEnv, config)
```
jonathan-mayer, line 122: we should handle the error here. and cancel the context if err != nil

cmd/kubedownscaler/main.go
```diff
@@ -80,15 +83,64 @@ func main() {
...
+			},
+			OnStoppedLeading: func() {
+				slog.Info("stopped leading")
```
jonathan-mayer, line 125: Im debating if we should implement this to actually stop the loop. Im guessing it would rely on this instantly stopping the application before further action. how it is right now, it would do nothing on stop and let run or die do its thing until it stops blocking and the main thread returns

cmd/kubedownscaler/main.go
jonathan-mayer: did you push the changes for this yet?

jonathan-mayer: Did you push the changes yet?

samuel-esp: sorry, I lost some part of the commit. Happens sometimes after running precommit, the memory/file system of my ide gets corrupted

cmd/kubedownscaler/main.go
samuel-esp: you mean like exiting from the application?

jonathan-mayer: yea, basically

cmd/kubedownscaler/main.go
```diff
@@ -142,8 +156,11 @@ func startScanning(client kubernetes.Client, ctx context.Context, layerCli, laye
 			"CliLayer", *layerCli,
 			"EnvLayer", *layerEnv,
 		)
-		os.Exit(1)
+
+		return err
```
jonathan-mayer, line 160: I wouldn't print the error here, instead return a fmt.Errorf formatted error

cmd/kubedownscaler/main.go
```diff
@@ -132,6 +132,7 @@ func main() {
 			},
 			OnStoppedLeading: func() {
 				slog.Info("stopped leading")
+				os.Exit(1)
```
jonathan-mayer, line 135: I thought more of a soft stop. like instead of doing while true implement while running (with context maybe) to stop the loop and cancel the context to stop the active api requests

cmd/kubedownscaler/main.go
```diff
@@ -157,7 +158,7 @@ func startScanning(
 			"EnvLayer", *layerEnv,
 		)
```
jonathan-mayer, line 159: would remove this slog.error

jonathan-mayer (changes requested):

cmd/kubedownscaler/main.go
samuel-esp: @jonathan-mayer could you check the new commit, I don't know if i understood you well. I don't know if you wanted me to create a new function just to stop the context. What I did was just to put the cancel() instead of the os.Exit(1)

jonathan-mayer:
> @jonathan-mayer before merging: what do you think about implemeting a custom argument to skip leader election if needed? like `--skip-leader-election`.
> 
> for example the [VPA](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/docs/faq.md#can-i-run-the-vpa-in-an-ha-configuration) lets the user choose if enabling the leader election or not

sure, lets then also update the helm chart to auto set it if replicas is >1 and make an additional value in the values.yaml to disable it

jonathan-mayer (changes requested):

cmd/kubedownscaler/main.go
jonathan-mayer: ive looked into this a bit more and this should be fine

cmd/kubedownscaler/main.go
```diff
@@ -80,18 +83,78 @@ func main() {
...
+		slog.Warn("couldn't get namespace or running outside of cluster; skipping leader election", "error", err)
+		slog.Warn("proceeding without leader election, this may cause multiple instances to conflict when modifying the same resources")
+
+		err = startScanning(client, ctx, &layerCli, &layerEnv, config)
+		if err != nil {
+			slog.Error("an error occurred while scanning workloads, exiting", "error", err)
+			os.Exit(1)
+		}
+
+		return
```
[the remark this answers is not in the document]
samuel-esp: @jonathan-mayer refactored, I saw the pipeline returned me a lint suggestion

cmd/kubedownscaler/main.go:34:1: calculated cyclomatic complexity for function main is 11, max is 10 (cyclop)

So I refactored the functions as showed in the last commit

cmd/kubedownscaler/main.go
```diff
@@ -32,35 +32,11 @@ const (
...
+	config := initializeConfig()
+	layerCli, layerEnv := initializeLayers()
 
-	config.ParseConfigFlags()
-
-	layerCli.ParseLayerFlags()
-
-	flag.Parse()
-
-	err := layerEnv.GetLayerFromEnv()
```
jonathan-mayer, line 58: i dont think these should be in functions in main, either we keep this all in the main func or we have to put these/refactor other functions in the packages responsible for handling these

cmd/kubedownscaler/main.go
```diff
@@ -88,17 +64,67 @@ func main() {
...
 	if err != nil {
 		slog.Warn("couldn't get namespace or running outside of cluster; skipping leader election", "error", err)
-		slog.Warn("proceeding without leader election, this may cause multiple instances to conflict when modifying the same resources")
-
-		err = startScanning(client, ctx, &layerCli, &layerEnv, config)
-		if err != nil {
-			slog.Error("an error occurred while scanning workloads, exiting", "error", err)
-			os.Exit(1)
-		}
+	}
```
jonathan-mayer, line 67: shouldnt it then actually skip leaderelection here, how it is right now unless leader election is also disabled it will just error because the namespace is "".

What we could also do is make leader election disabled by default, only enabled via the argument. Put this getting the namespace after the check if leader election is disabled and error and exit when the namespace couldn't be gotten.

internal/pkg/util/config.go
```diff
@@ -14,6 +14,8 @@ type RuntimeConfiguration struct {
 	Debug bool
 	// Once sets if the scan should only run once.
 	Once bool
+	// LeaderElectionEnabled sets if leader election should be performed.
+	LeaderElectionEnabled bool
```
jonathan-mayer, line 18:
```suggestion
	// LeaderElection sets if leader election should be performed.
	LeaderElection bool
```

internal/pkg/util/config.go
```diff
@@ -52,6 +54,12 @@ func (c *RuntimeConfiguration) ParseConfigFlags() {
 		false,
 		"run scan only once (default: false)",
 	)
+	flag.BoolVar(
+		&c.Once,
+		"leader-election-enabled",
+		false,
+		"leader election is not enabled (default: false)",
+	)
```
jonathan-mayer, line 62:
```suggestion
	flag.BoolVar(
		&c.Once,
		"leader-election",
		false,
		"enables leader election (default: false)",
	)
```

deployments/chart/templates/leaserole.yaml
jonathan-mayer, line 1: put a if around all this to only make the role when leader election is enabled. for this i would add a leader election enabled field in the helpers and use it here and in the deployment.yaml where you add the cli argument to enable leader election

jonathan-mayer (changes requested):

cmd/kubedownscaler/main.go
```diff
@@ -61,60 +90,18 @@ func main() {
...
-	return layerCli, layerEnv
-}
-
-func setupEnv(layerEnv *values.Layer) error {
-	err := layerEnv.GetLayerFromEnv()
+	downscalerNamespace, err := kubernetes.GetCurrentNamespace()
 	if err != nil {
-		return fmt.Errorf("failed to get layer from env: %w", err)
+		slog.Warn("couldn't get namespace or running outside of cluster; skipping leader election", "error", err)
+		runWithoutLeaderElection(client, ctx, &layerCli, &layerEnv, config)
```
jonathan-mayer, line 101:
```suggestion
		slog.Error("failed to get namespace or running outside of cluster; disable leader election to continue", "error", err)
		os.Exit(1)
```

jonathan-mayer (changes requested):

samuel-esp:
Refactored the helm chart, left open one discussion above

deployments/chart/templates/_helpers.tpl
```diff
@@ -24,10 +24,10 @@ Create chart name and version as used by the chart label.
 {{- end }}
 
 {{/*
-Check if replicaCount > 1 or if leader election is enabled. If replicaCount is 1, leader election is overridden to true.
+If replicaCount > leader election is enabled by default.
```
jonathan-mayer, line 27:
```suggestion
If replicaCount is greater than 1 leader election is enabled by default.
```

deployments/chart/values.yaml
```diff
@@ -1,4 +1,5 @@
 replicaCount: 1
+# If replicaCount is greater than 1, the leader election is enabled by default
```
jonathan-mayer, line 2:
```suggestion
# If replicaCount is greater than 1, the leader election is enabled by default
replicaCount: 1
```

cmd/kubedownscaler/main.go
jonathan-mayer: Just to reiterate from my previous comment on this. I think it would be good to if someone explicitly enables leader election via the cli argument, we should enforce that wish. This could otherwise easily lead to unwanted behavior from wrong configuration. One of the main improvements of the gokubedownscaler is that unclear/wrong configuration will throw an error to show the user their fault, instead of just ignoring it and doing unexpected things.

cmd/kubedownscaler/main.go
```diff
@@ -80,20 +88,103 @@ func main() {
...
+	if !config.LeaderElection {
+		runWithoutLeaderElection(client, ctx, &layerCli, &layerEnv, config)
+		return
+	}
+
+	downscalerNamespace, err := kubernetes.GetCurrentNamespace()
+	if err != nil {
+		slog.Warn("couldn't get namespace or running outside of cluster; skipping leader election", "error", err)
+		os.Exit(1)
+	}
```
jonathan-mayer, line 102: we can put this into the client.CreateLease function and with that make the func private

cmd/kubedownscaler/main.go
```diff
@@ -80,20 +88,103 @@ func main() {
...
+	client kubernetes.Client,
+	downscalerNamespace string,
+	cancel context.CancelFunc,
+	ctx context.Context,
+	layerCli, layerEnv *values.Layer,
+	config *util.RuntimeConfiguration,
+) {
+	lease, err := client.CreateLease(leaseName, downscalerNamespace)
+	if err != nil {
+		slog.Warn("failed to create lease", "error", err)
```
jonathan-mayer, line 117:
```suggestion
		slog.Error("failed to create lease", "error", err)
```

cmd/kubedownscaler/main.go
```diff
@@ -80,20 +88,103 @@ func main() {
...
+	ctx, cancel := context.WithCancel(context.Background())
+
+	if !config.LeaderElection {
+		runWithoutLeaderElection(client, ctx, &layerCli, &layerEnv, config)
+		return
+	}
+
+	downscalerNamespace, err := kubernetes.GetCurrentNamespace()
+	if err != nil {
+		slog.Warn("couldn't get namespace or running outside of cluster; skipping leader election", "error", err)
```
jonathan-mayer, line 100:
```suggestion
		slog.Error("couldn't get namespace or running outside of cluster", "error", err)
```

cmd/kubedownscaler/main.go
```diff
@@ -80,20 +88,103 @@ func main() {
...
+	})
+}
+
+func runWithoutLeaderElection(
+	client kubernetes.Client,
+	ctx context.Context,
+	layerCli, layerEnv *values.Layer,
+	config *util.RuntimeConfiguration,
+) {
+	slog.Warn("proceeding without leader election, this may cause multiple downscaler instances to conflict when modifying the same resources")
```
jonathan-mayer, line 163:
```suggestion
	slog.Warn("proceeding without leader election; this could cause errors when running with multiple replicas")
```

cmd/kubedownscaler/main.go
```diff
@@ -80,20 +88,103 @@ func main() {
...
+	if err != nil {
+		return fmt.Errorf("failed to scan over workloads: %w", err)
+	}
+
+	return nil
+}
+
 // scanWorkloads scans over all workloads every scan.
 func scanWorkloads(
 	client kubernetes.Client,
```
jonathan-mayer, line 190: could we actually combine the start Scanning and scanWorkloads function, it only adds 1 log line and removes a bit of error handling boiler plate

deployments/chart/templates/leaserole.yaml
```diff
@@ -0,0 +1,32 @@
+{{- if include "go-kube-downscaler.leaderElection" . }}
```
jonathan-mayer, line 1:
```suggestion
{{- if include "go-kube-downscaler.leaderElection" . -}}
```

deployments/chart/values.yaml
```diff
@@ -1,3 +1,4 @@
...
```
jonathan-mayer, line 1:
```suggestion
# If replicaCount is greater than 1, leader election is enabled by default
```

jonathan-mayer (changes requested):
Ok, im pretty sure we can merge after this

internal/api/kubernetes/client.go
```diff
@@ -245,13 +245,19 @@ func (c client) addWorkloadEvent(eventType, reason, identifier, message string,
...
 	if err != nil {
 		slog.Error("failed to get hostname", "error", err)
 		return nil, fmt.Errorf("failed to get hostname: %w", err)
 	}
 
+	leaseNamespace, err := getCurrentNamespace()
+	if err != nil {
+		slog.Error("couldn't get namespace or running outside of cluster", "error", err)
+		os.Exit(1)
+	}
```
jonathan-mayer, line 259: return error here

jonathan-mayer (changes requested):

internal/api/kubernetes/client.go
```diff
@@ -240,3 +244,30 @@ func (c client) addWorkloadEvent(eventType, reason, identifier, message string,
...
+	if err != nil {
+		slog.Error("failed to get hostname", "error", err)
+		return nil, fmt.Errorf("failed to get hostname: %w", err)
+	}
+
+	leaseNamespace, err := getCurrentNamespace()
+	if err != nil {
+		slog.Error("couldn't get namespace or running outside of cluster", "error", err)
+		return nil, fmt.Errorf("couldn't get namespace: %w", err)
+	}
```
jonathan-mayer, line 259:
```suggestion
	leaseNamespace, err := getCurrentNamespace()
	if err != nil {
		return nil, fmt.Errorf("failed to get namespace or running outside of cluster: %w", err)
	}
```

cmd/kubedownscaler/main.go
```diff
@@ -80,27 +88,88 @@ func main() {
...
+	config *util.RuntimeConfiguration,
+) {
+	lease, err := client.CreateLease(leaseName)
+	if err != nil {
+		slog.Error("failed to create lease", "error", err)
+		os.Exit(1)
+	}
+
+	defer cancel()
+
```
jonathan-mayer, line 115: please put this directly after initializing the context

internal/api/kubernetes/client.go
```diff
@@ -240,3 +244,29 @@ func (c client) addWorkloadEvent(eventType, reason, identifier, message string,
 
 	return nil
 }
+
+func (c client) CreateLease(leaseName string) (*resourcelock.LeaseLock, error) {
+	hostname, err := os.Hostname()
+	if err != nil {
+		slog.Error("failed to get hostname", "error", err)
```
jonathan-mayer, line 251:
```suggestion
```

jonathan-mayer (changes requested):

jonathan-mayer (approved):

jonathan-mayer:
i totally forgot to merge this

Issue #64, opened 2024-10-16
broken ScaledObjects scaling


jonathan-mayer:
## Issue

Workloads managed by ScaledObjects are currently also scaled by the Downscaler. This results in the ScaledObject and the downscaler fighting to downscale those workloads to (possibly) different Replicas.

### Further details

Bug was found here: https://github.com/caas-team/py-kube-downscaler/issues/96
Bug was fixed in py-kube-downscaler here: https://github.com/caas-team/py-kube-downscaler/pull/100

### Proposal

Implement some way to avoid this.
A possible way would be to automatically exclude the workloads managed by ScaledObjects from being scaled by the Downscaler.

samuel-esp:
My suggestion would be something like:

```go
// GetWorkloads gets all workloads of the specified resources for the specified namespaces
func (c client) GetWorkloads(namespaces []string, resourceTypes []string, ctx context.Context) ([]scalable.Workload, error) {
	var results []scalable.Workload
	if namespaces == nil {
		namespaces = append(namespaces, "")
	}
	for _, namespace := range namespaces {
		for _, resourceType := range resourceTypes {
			slog.Debug("getting workloads from resource type", "resourceType", resourceType)
			getWorkloads, ok := scalable.GetWorkloads[strings.ToLower(resourceType)]
			if !ok {
				return nil, errResourceNotSupported
			}
			workloads, err := getWorkloads(namespace, c.clientsets, ctx)
			if err != nil {
				return nil, fmt.Errorf("failed to get workloads: %w", err)
			}
			results = append(results, workloads...)
		}
	}

	return results, nil
}
```

- Insert a conditional for this case:
  - When ScaledObjects are retrieved, iterate over each object and insert the Deployment/StatefulSet reference (name) inside an HashMap
- Insert a conditional for this case:
  - When Deployments are retrieved, iterate over each object. They are not appended to the results list if the name is inside the HashMap
  - When StatefulSets are retrieved, iterate over each object.  They are not appended to the results list if the name is inside the HashMap

This approach requires the ScaledObjects to be processed before Deployments/StatefulSets

Things to consider: clarity of code (right now is perfect), performance (should still be 0(N), but we will iterate many times inside lists that could be pretty big, especially the Deployment list)

jonathan-mayer:
I'm not sure if putting all this in the GetWorkloads function is right. I will have a look on how to implement it next week as I am still out of Office for the rest of this Week.

PR #76, opened 2024-12-21
Feat/exclude workloads managed by scaled object

samuel-esp:
## Motivation

As stated in #64: workloads managed by ScaledObjects are currently also scaled by the Downscaler. This results in the ScaledObject and the downscaler fighting to downscale those workloads to (possibly) different Replicas.

This PRs solves this problem by excluding from downscaling all the workloads managed by a Keda ScaledObject. The solution works in this way:

**_Prerequisite_**: all ScaledObjects should be processed first by the algorithm 

- Inside the `FilterExcluded` function:
  - When the algorithm encounters a ScaledObject:
    - It computes a hash for the ScaledObject's `targetRef`, which is derived from the combination of (Kind+Name+Namespace):
    - The computed hash is stored in a hashMap for reference.
  - When the algorithm encounters every other workloads (Deployment, StatefulSet, ...):
    - The `isManagedByKeda` function computes a hash for the workload using (Kind+Name+Namespace)
    - It checks if the computed hash exists in the hashMap.
      - If a match is found in the hashMap, the workload is excluded from scaling as it is considered managed by KEDA. 

## Changes

- Refactored `FilterExcluded` to implement the solution proposed
- Introduced `computeHash` function to compute a Sum64 hash
- Introduce `OrderIncludeResourcesArgument` function to enforce the order of submission for resources inside `--include-resources` argument
- Introduced inside each Kubernetes Native class its `GroupVersionKind`

## Tests done

- Unit Tests
- Live Tests

## TODO

- [ ] Check Native GroupVersionKind Existence for these CRDs (Rollout, Stack, Prometheus)
- [ ] Add More Unit Tests

### DIFF
--- a/cmd/kubedownscaler/main.go
+++ b/cmd/kubedownscaler/main.go
@@ -97,6 +97,7 @@ func main() {
 		slog.Error("found incompatible fields", "error", err)
 		os.Exit(1)
 	}
+	includeResources = values.OrderIncludeResourcesArgument(includeResources)
 	ctx := context.Background()
 
 	slog.Debug("getting client for kubernetes")
@@ -109,7 +110,6 @@ func main() {
 	slog.Info("started downscaler")
 	for {
 		slog.Info("scanning workloads")
-
 		workloads, err := client.GetWorkloads(includeNamespaces, includeResources, ctx)
 		if err != nil {
 			slog.Error("failed to get workloads", "error", err)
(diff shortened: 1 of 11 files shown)


jonathan-mayer:
Ive took some inspiration from this and refactored it in #78 to be a bit less verbose and fit nicer in the preexisting structure. Nice work @samuel-esp.

samuel-esp:
That's perfect Jonathan, don't worry!

PR #77, opened 2025-01-03
fix: make workflows run for forks

JTaeuber:
## Motivation
Our workflows didn't work for forks.

## Changes
Changed type to pull_request_target.

## Tests done
Tested in a repository.


jonathan-mayer (approved):
no hax

PR #78, opened 2025-01-03
Feat/exclude externally scaled workloads

### LINKED ISSUE #64: broken ScaledObjects scaling

jonathan-mayer:
## Motivation

Closes #64 

## Tests done

- [X] unit tests passed


JTaeuber (approved):

internal/pkg/scalable/util.go
```diff
@@ -82,47 +91,56 @@ func getExternallyScaled(workloads []Workload) []workloadIdentifier {
 			namespace: scaledobject.Namespace,
 		})
 	}
-	return externallyScaled
+
+	return externallyScaled[:len(externallyScaled):len(externallyScaled)]
```
JTaeuber, line 95: Do we leave this as is now or should we add a comment here as well?

PR #81, opened 2025-01-09
perf: reduce memory allocations

jonathan-mayer:
## Motivation

This PR aims to reduce memory reallocations by refactoring for loops which append to slices to resize the underlying array where ever possible.
Additionally it removes unnecessary temporary copies in the get resource functions.

This is in theory slightly going to improve performance and efficiency. It wouldn't hurt to go over the whole project and try to find more places where memory allocations could be avoided. That's a lot of work though and its not like the downscaler has performance issues so it would just be a nice to have.

## Tests done

- [X] unit tests passed


JTaeuber (approved):
Sure ¯\\\_(ツ)\_/¯

PR #100, opened 2025-01-27
Refactor/logs by adding stringers

samuel-esp:
## Motivation

This is just a small PR to fix some minor issues

## Changes

1. Refactored some logs in order to print the real string value and not pointers
2. Added support to parse "EXCLUDE_NAMESPACES" and "EXCLUDE_DEPLOYMENTS" from Environment Variables

## Tests done

- Unit tests

### DIFF
--- a/cmd/kubedownscaler/main.go
+++ b/cmd/kubedownscaler/main.go
@@ -186,9 +186,8 @@ func scanWorkload(
 	}
 
 	layers := values.Layers{&layerWorkload, &layerNamespace, layerCli, layerEnv}
-	layersString := layers.LayersToString(layers)
 
-	slog.Debug("finished parsing all layers", "layers", layersString, "workload", workload.GetName(), "namespace", workload.GetNamespace())
+	slog.Debug("finished parsing all layers", "layers", layers, "workload", workload.GetName(), "namespace", workload.GetNamespace())
 
 	isInGracePeriod, err := layers.IsInGracePeriod(
 		config.TimeAnnotation,
--- a/cmd/kubedownscaler/main.go
+++ b/cmd/kubedownscaler/main.go
@@ -186,8 +186,9 @@ func scanWorkload(
 	}
 
 	layers := values.Layers{&layerWorkload, &layerNamespace, layerCli, layerEnv}
+	layersString := layers.LayersToString(layers)
 
-	slog.Debug("finished parsing all layers", "layers", layers, "workload", workload.GetName(), "namespace", workload.GetNamespace())
+	slog.Debug("finished parsing all layers", "layers", layersString, "workload", workload.GetName(), "namespace", workload.GetNamespace())
 
 	isInGracePeriod, err := layers.IsInGracePeriod(
 		config.TimeAnnotation,
--- a/internal/pkg/values/layer.go
+++ b/internal/pkg/values/layer.go
@@ -264,12 +264,12 @@ func (l Layers) String() string {
 
 	builder.WriteString("[")
 
-	for iteration, item := range l {
-		if iteration > 0 {
+	for i, layer := range l {
+		if i > 0 {
 			builder.WriteString(" ")
 		}
 
-		fmt.Fprintf(&builder, "%s:%+v", LayerID(iteration), item)
+		fmt.Fprintf(&builder, "%s:%+v", LayerID(i), layer)
 	}
 
 	builder.WriteString("]")
--- a/internal/pkg/values/layer.go
+++ b/internal/pkg/values/layer.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strconv"
 	"strings"
 	"time"
 
@@ -18,11 +17,6 @@ var (
 	errInvalidDownscaleReplicas = errors.New("error: downscale replicas value is invalid")
 	errValueNotSet              = errors.New("error: no layer implements this value")
 	errAnnotationNotSet         = errors.New("error: annotation isn't set on workload")
-)
-
-const (
-	Undefined       = -1          // Undefined represents an undefined integer value
-	UndefinedString = "undefined" // UndefinedString epresents an undefined value
 )
 
 // Scaling is an enum that describes the current Scaling.
@@ -58,8 +52,8 @@ func (l LayerID) String() string {
 // NewLayer gets a new layer with the default values.
 func NewLayer() Layer {
 	return Layer{
-		DownscaleReplicas: Undefined,
-		GracePeriod:       Undefined,
+		DownscaleReplicas: util.Undefined,
+		GracePeriod:       util.Undefined,
 	}
 }
 
@@ -100,7 +94,7 @@ func (l *Layer) CheckForIncompatibleFields() error { //nolint: cyclop // this is
 		return errForceUpAndDownTime
 	}
 	// downscale replicas invalid
-	if l.DownscaleReplicas != Undefined && l.DownscaleReplicas < 0 {
+	if l.DownscaleReplicas != util.Undefined && l.DownscaleReplicas < 0 {
 		return errInvalidDownscaleReplicas
 	}
 	// up- and downtime
@@ -194,7 +188,7 @@ func (l Layers) GetCurrentScaling() Scaling {
 func (l Layers) GetDownscaleReplicas() (int32, error) {
 	for _, layer := range l {
 		downscaleReplicas := layer.DownscaleReplicas
-		if downscaleReplicas == Undefined {
+		if downscaleReplicas == util.Undefined {
 			continue
 		}
 
@@ -227,10 +221,10 @@ func (l Layers) IsInGracePeriod(
 	ctx context.Context,
 ) (bool, error) {
 	var err error
-	var gracePeriod time.Duration = Undefined
+	var gracePeriod time.Duration = util.Undefined
 
 	for _, layer := range l {
-		if layer.GracePeriod == Undefined {
+		if layer.GracePeriod == util.Undefined {
 			continue
 		}
 
@@ -239,7 +233,7 @@ func (l Layers) IsInGracePeriod(
 		break
 	}
 
-	if gracePeriod == Undefined {
+	if gracePeriod == util.Undefined {
 		return false, nil
 	}
 
@@ -275,7 +269,7 @@ func (l Layers) String() string {
 			builder.WriteString(" ")
 		}
 
-		fmt.Fprintf(&builder, "%s:%+v", LayerId(i), item)
+		fmt.Fprintf(&builder, "%s:%+v", LayerID(iteration), item)
 	}
 
 	builder.WriteString("]")
--- a/internal/pkg/values/layer.go
+++ b/internal/pkg/values/layer.go
@@ -18,7 +18,10 @@ var (
 	errAnnotationNotSet         = errors.New("error: annotation isn't set on workload")
 )
 
-const Undefined = -1 // Undefined represents an undefined integer value
+const (
+	Undefined       = -1          // Undefined represents an undefined integer value
+	UndefinedString = "undefined" // UndefinedString epresents an undefined value
+)
 
 // Scaling is an enum that describes the current Scaling.
 type Scaling int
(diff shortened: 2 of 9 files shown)


internal/pkg/values/timespan.go
```diff
@@ -75,7 +75,38 @@ func (t *timeSpans) Set(value string) error {
...
+		default:
+			stringBuilder.WriteString("UnknownTimeSpan")
+		}
+	}
+
+	if stringBuilder.Len() == 0 {
+		return UndefinedString
+	}
+
+	return stringBuilder.String()
```
jonathan-mayer, line 109: This fmt.Stringer implementation shouldn't be on the timeSpans type. Instead it should be on the individual implementations of the TimeSpan interface.

internal/pkg/values/timespan.go
```diff
@@ -75,7 +75,38 @@ func (t *timeSpans) Set(value string) error {
...
+	var stringBuilder strings.Builder
+
+	for i, timespan := range *t {
+		if i > 0 {
+			stringBuilder.WriteString(", ")
+		}
+
+		switch span := timespan.(type) {
+		case absoluteTimeSpan:
+			stringBuilder.WriteString(fmt.Sprintf("AbsoluteTimeSpan(%s - %s)", span.from.Format(time.RFC3339), span.to.Format(time.RFC3339)))
```
jonathan-mayer, line 87: i would change `AbsoluteTimeSpan` to the actual type name, so `absoluteTimeSpan`
```suggestion
			stringBuilder.WriteString(fmt.Sprintf("absoluteTimeSpan(%s - %s)", span.from.Format(time.RFC3339), span.to.Format(time.RFC3339)))
```

internal/pkg/values/timespan.go
```diff
@@ -75,7 +75,38 @@ func (t *timeSpans) Set(value string) error {
...
+			weekdayTo := span.weekdayTo.String()[:3]     // Get first 3 letters of the weekday
+			timeFrom := span.timeFrom.Format("15:04")    // Format time to "HH:MM"
+			timeTo := span.timeTo.Format("15:04")        // Format time to "HH:MM"
+			stringBuilder.WriteString(fmt.Sprintf(
+				"RelativeTimeSpan(%s-%s %s-%s %s)",
+				weekdayFrom,
+				weekdayTo,
+				timeFrom,
+				timeTo,
+				span.timezone.String()))
```
jonathan-mayer, line 99: we shouldnt allocate all of these values. just use them directly.
```suggestion
			weekdayFrom := span.weekdayFrom.String()[:3] // Get first 3 letters of the weekday
			weekdayTo := span.weekdayTo.String()[:3]     // Get first 3 letters of the weekday
			timeFrom := span.timeFrom.Format("15:04")    // Format time to "HH:MM"
			timeTo := span.timeTo.Format("15:04")        // Format time to "HH:MM"
			stringBuilder.WriteString(fmt.Sprintf(
				"relativeTimeSpan(%.3s-%.3s %s-%s %s)",
				span.weekdayFrom,
				span.weekdayTo,
				span.timeFrom.Format("15:04"),
				span.timeTo.Format("15:04"),
				span.timezone))
```

internal/pkg/values/layer.go
```diff
@@ -238,3 +243,48 @@ func (l Layers) IsInGracePeriod(
...
+			layer.ForceDowntime.String(),
+			layer.DownscaleReplicas,
+			layer.GracePeriod,
+			layer.ExcludeNamespaces,
+			layer.ExcludeDeployments,
+		)
+	}
+
+	return result
+}
```
jonathan-mayer, line 290: I dont see how having these layer names hard coded like this is any good. 
I think its good to have them somewhat like this in an enum.

```go
type LayerId int

const (
	LayerWorkload LayerId = iota
	LayerNamespace
	LayerCli
	LayerEnvironment
)

func (l LayerId) String() string {
	return map[LayerId]string{
		LayerWorkload:    "LayerWorkload",
		LayerNamespace:   "LayerNamespace",
		LayerCli:         "LayerCli",
		LayerEnvironment: "LayerEnvironment",
	}[l]
}
```

Then we can just do this. This also removed the unnecessarily verbose `.String()` syntax and uses the `%+v` format syntax to make showing the Names of the Fields less verbose. Only difference to yours should be that yours has new lines, this one doesn't.

```suggestion
func (l Layers) String() string {
	var builder strings.Builder
	builder.WriteString("[")

	for i, item := range l {
		if i > 0 {
			builder.WriteString(" ")
		}
		fmt.Fprintf(&builder, "%s:%+v", LayerId(i), item)
	}

	builder.WriteString("]")
	return builder.String()
}
```

Along with this i would also make an array out of the Layer slice type, so:
```go
type Layers [4]*Layer
```

this should be a simple drop in replacement since we only use the Layers type with 4 arguments

jonathan-mayer (changes requested):

internal/pkg/values/layerParser.go
```diff
@@ -69,6 +71,22 @@ func (l *Layer) ParseLayerFlags() {
...
+		)
+	}
+
+	if flag.Lookup("exclude-deployments") == nil {
+		flag.Var(
+			&l.ExcludeDeployments,
+			"exclude-deployments",
+			"exclude deployments from being scaled (optional)",
+		)
+	}
```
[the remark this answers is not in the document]
samuel-esp: Unfortunately I don't think we can avoid `flag.Lookup` since it will cause a panic error otherwise (flag is redefined). From what i tested, the `&l.ExcludedNamespaces` `&l.ExcludedDeployments` is called inside 2 methods and this is the reason the panic error is raised. The only way to avoid this is to check if the value is already present 

 (https://stackoverflow.com/questions/49193480/golang-flag-redefined)

cmd/kubedownscaler/main.go
```diff
@@ -186,8 +186,9 @@ func scanWorkload(
 	}
 
 	layers := values.Layers{&layerWorkload, &layerNamespace, layerCli, layerEnv}
+	layersString := layers.LayersToString(layers)
 
-	slog.Debug("finished parsing all layers", "layers", layers, "workload", workload.GetName(), "namespace", workload.GetNamespace())
+	slog.Debug("finished parsing all layers", "layers", layersString, "workload", workload.GetName(), "namespace", workload.GetNamespace())
```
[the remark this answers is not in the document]
samuel-esp: I'll resolve this asap

internal/pkg/values/layer.go
samuel-esp: Sorry my implementation was horrendous, I'll refactor in this way

internal/pkg/values/layerParser.go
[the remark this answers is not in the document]
samuel-esp: just to align others that will read the topic in the future, there was a little misunderstanding here that came from the fact that the parsing of EXCLUDE_NAMESPACES and EXCLUDED_DEPLOYMENTS env vars was accidentally removed in the latest refactoring merged on main (commit [f6dbfc0](https://github.com/caas-team/GoKubeDownscaler/commit/f6dbfc0b49c99de0f90c01230b7f02195028e732)). I tried to changed this portion of code because I thought the overall code structure for parsing was changed. I addressed this to Jonhatan and he found out what was missing in #101

internal/pkg/values/layer.go
```diff
@@ -40,16 +43,18 @@ func NewLayer() Layer {
...
+	UpscalePeriod      timeSpans      // periods to upscale in
+	UpTime             timeSpans      // within these timespans workloads will be scaled up, outside of them they will be scaled down
+	Exclude            triStateBool   // if workload should be excluded
+	ExcludeUntil       time.Time      // until when the workload should be excluded
+	ForceUptime        triStateBool   // force workload into a uptime state
+	ForceDowntime      triStateBool   // force workload into a downtime state
+	DownscaleReplicas  int32          // the replicas to scale down to
+	GracePeriod        time.Duration  // grace period until new workloads will be scaled down
+	ExcludeNamespaces  util.RegexList // excluded namespaces
+	ExcludeDeployments util.RegexList // excluded deployments
```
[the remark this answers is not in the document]
samuel-esp: As i said in the previous conversation

> just to align others that will read the topic in the future, there was a little misunderstanding here that came from the fact that the parsing of EXCLUDE_NAMESPACES and EXCLUDED_DEPLOYMENTS env vars was accidentally removed in the latest refactoring merged on main (commit [f6dbfc0](https://github.com/caas-team/GoKubeDownscaler/commit/f6dbfc0b49c99de0f90c01230b7f02195028e732)). I tried to changed this portion of code because I thought the overall code structure for parsing was changed. I addressed this to Jonhatan and he found out what was missing in https://github.com/caas-team/GoKubeDownscaler/pull/101
> 
> I will revert this change asap

samuel-esp:
Just refactored according the suggestion, thanks for reviewing it Jonathan!

internal/pkg/values/layer.go
```diff
@@ -244,47 +264,23 @@ func (l Layers) IsInGracePeriod(
...
-			layer.ForceUptime.String(),
-			layer.ForceDowntime.String(),
-			layer.DownscaleReplicas,
-			layer.GracePeriod,
-			layer.ExcludeNamespaces,
-			layer.ExcludeDeployments,
-		)
+		builder.WriteString(strconv.Itoa(int(LayerID(iteration))))
+		builder.WriteString(":")
+		builder.WriteString(fmt.Sprintf("%+v", item))
```
jonathan-mayer, line 280: I think if we are already using fmt.Sprintf we could just put the rest in there too. Also at least theoretically fmt.Fprintf(builder) is slightly more efficient.
```suggestion
		fmt.Fprintf(&builder, "%s:%+v", LayerId(i), item)
```

internal/pkg/values/layer.go
```diff
@@ -244,47 +264,23 @@ func (l Layers) IsInGracePeriod(
...
-			layer.Exclude.String(),
-			excludeUntilStr,
-			layer.ForceUptime.String(),
-			layer.ForceDowntime.String(),
-			layer.DownscaleReplicas,
-			layer.GracePeriod,
-			layer.ExcludeNamespaces,
-			layer.ExcludeDeployments,
-		)
+		builder.WriteString(strconv.Itoa(int(LayerID(iteration))))
```
jonathan-mayer, line 278: Won't this just print it as a base 10 integer representation string? We have a Stringer on the LayerID for a reason.

internal/pkg/values/timespan.go
```diff
@@ -41,6 +41,32 @@ func (t *timeSpans) inTimeSpans() bool {
...
+		case *relativeTimeSpan:
+			builder.WriteString(span.String())
+		default:
+			builder.WriteString("UnknownTimeSpan")
+		}
+	}
+
+	return builder.String()
+}
+
```
jonathan-mayer, line 69: We shouldn't need this stringer at all. Golang should already format this just like a normal array.

internal/pkg/values/timespan.go
```diff
@@ -223,6 +226,11 @@ func (t absoluteTimeSpan) isTimeInSpan(targetTime time.Time) bool {
 	return (t.from.Before(targetTime) || t.from.Equal(targetTime)) && t.to.After(targetTime)
 }
 
+// String implementation for absoluteTimeSpan.
+func (t absoluteTimeSpan) String() string {
+	return fmt.Sprintf("absoluteTimeSpan(%s - %s)", t.from.Format(time.RFC3339), t.to.Format(time.RFC3339))
```
jonathan-mayer, line 231:
```suggestion
	return fmt.Sprintf(
		"absoluteTimeSpan(%s - %s)", 
		t.from.Format(time.RFC3339),
		t.to.Format(time.RFC3339),
	)
```

jonathan-mayer (changes requested):

jonathan-mayer (changes requested):

internal/pkg/values/timespan.go
```diff
@@ -74,10 +100,6 @@ func (t *timeSpans) Set(value string) error {
 	return nil
 }
 
-func (t *timeSpans) String() string {
-	return fmt.Sprint(*t)
-}
-
```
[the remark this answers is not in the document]
samuel-esp: fixed

internal/pkg/values/layer.go
```diff
@@ -18,7 +20,10 @@ var (
 	errAnnotationNotSet         = errors.New("error: annotation isn't set on workload")
 )
 
-const Undefined = -1 // Undefined represents an undefined integer value
+const (
+	Undefined       = -1          // Undefined represents an undefined integer value
+	UndefinedString = "undefined" // UndefinedString epresents an undefined value
+)
```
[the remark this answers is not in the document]
samuel-esp: moved to util.go class and refactored all the scalable methods that were referencing "`Undefined`"

internal/pkg/values/timespan.go
```diff
@@ -43,28 +43,11 @@ func (t *timeSpans) inTimeSpans() bool {
 
 // String implementation for timeSpans.
 func (t *timeSpans) String() string {
-	if len(*t) == 0 {
-		return UndefinedString
+	if *t != nil {
+		return "invalid value"
```
jonathan-mayer, line 47: we should probably do `nil` here or `timeSpans(nil)` or maybe just use the undefinedString const

jonathan-mayer (changes requested):
I think after this we are ready to merge this.

samuel-esp:
Solved

internal/pkg/values/layer.go
```diff
@@ -238,3 +257,22 @@ func (l Layers) IsInGracePeriod(
...
 	return time.Now().Before(gracePeriodUntil), nil
 }
+
+// String gets the string representation of the layers.
+func (l Layers) String() string {
+	var builder strings.Builder
+
+	builder.WriteString("[")
+
+	for iteration, item := range l {
```
jonathan-mayer, line 267: rename to `i` and `layer`

jonathan-mayer (changes requested):

internal/pkg/values/layer.go
samuel-esp: @jonathan-mayer ok for `layer`, the `iteration` variable was called `i` previously but i had to rename it because the linter blocked me in the pre-commit. Should we go for nolint on this?

jonathan-mayer: it really shouldnt error because of the i, can you try again

samuel-esp: done, probably the error i encountered was just because of another lint suggestion

jonathan-mayer (approved):

PR #101, opened 2025-01-27
fix: env vars parsing for runtime config

jonathan-mayer:
## Motivation

Env vars were not being parsed after the last refactoring. This fixes this and additionally slightly refactors the order of parsing env and flags.

Bug found by @samuel-esp.

### DIFF
--- a/cmd/kubedownscaler/main.go
+++ b/cmd/kubedownscaler/main.go
@@ -42,24 +42,29 @@ func main() {
 		Kubeconfig:        "",
 	}
 
+	config.ParseConfigFlags()
+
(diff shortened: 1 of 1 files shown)


samuel-esp:
I found out this bug accidentally when testing the images for #75 and #100 , as mentioned in the conversations from #100 I thought the code structure for parsing these env variables changed and they were forget in the process, but in reality that was not the case because there was a function in the code that was never called (`ParseConfigEnvVars`) which wasn't a leftover from the old code structure

JTaeuber (approved):
👍

Issue #110, opened 2025-02-12
Refactor relative timespans


jonathan-mayer:
## Issue

Currently relative timespans use "zero time" a imaginary point in time which maps to the 1st of January 0. 
The main reason this was used is to avoid implementing a own implementation of time for just the time of day.

### Problem to solve

This is hard to understand and a bit overkill.

### Proposal

Refactor the RelativeTimeSpan to use a custom day time implementation with the necessary methods.

### Who can address the issue

everyone with go experience

PR #118, opened 2025-02-20
Refactor/timespans and layer defaults

### LINKED ISSUE #117: Refactor to use default layer
## Issue

Currently layer defaults are set on the cli layer, this is just dumb. Especially since the env layer then cant overwrite those and setting the defaults is hard with flag.Var since it doesn't support defaults and they have to be set manually.

### Proposal

- [ ] Add a new layer, the "default" layer. In that layer, all defaults should be defined. The default states for all other layers should be the unset values. The default layer is the lowest layer. 
- [ ] Create a MR/modify feat/add-content-to-site to update the docs on layers, including the explanatory image.

### Who can address the issue

Anyone

jonathan-mayer:
## Motivation

Closes #117 and #110

### DIFF
--- a/cmd/kubedownscaler/main.go
+++ b/cmd/kubedownscaler/main.go
@@ -7,7 +7,6 @@ import (
 	"log/slog"
 	"os"
 	"os/signal"
-	"regexp"
 	"sync"
 	"syscall"
 	"time"
@@ -21,32 +20,11 @@ import (
 )
 
 const (
-	// value defaults.
-	defaultGracePeriod       = 15 * time.Minute
-	defaultDownscaleReplicas = 0
-
 	leaseName = "downscaler-lease"
-
-	// runtime config defaults.
-	defaultInterval = 30 * time.Second
 )
 
 func main() {
-	// set defaults for runtime configuration
-	config := &util.RuntimeConfiguration{
-		DryRun:            false,
-		Debug:             false,
-		Once:              false,
-		Interval:          defaultInterval,
-		IncludeNamespaces: nil,
-		IncludeResources:  []string{"deployments"},
-		ExcludeNamespaces: util.RegexList{regexp.MustCompile("kube-system"), regexp.MustCompile("kube-downscaler")},
-		ExcludeWorkloads:  nil,
-		IncludeLabels:     nil,
-		TimeAnnotation:    "",
-		Kubeconfig:        "",
-	}
-
+	config := util.GetDefaultConfig()
 	config.ParseConfigFlags()
 
 	err := config.ParseConfigEnvVars()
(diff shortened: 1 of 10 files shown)


internal/pkg/values/dayTime.go
```diff
@@ -0,0 +1,60 @@
...
+	"strings"
+	"time"
+)
+
+const (
+	Hour   dayTime = 60
+	Minute dayTime = 1
+)
+
+var errTimeOfDayOutOfRange = errors.New("error: the specified time of day is out of range")
```
JTaeuber, line 16: Shouldn't this also be a constant?

internal/pkg/values/timespan.go
```diff
@@ -172,18 +174,18 @@ func (t relativeTimeSpan) isWeekdayInRange(weekday time.Weekday) bool {
...
-func (t relativeTimeSpan) isTimeOfDayInRange(timeOfDay time.Time) bool {
-	if t.timeFrom.After(t.timeTo) { // check if range wraps across days
-		return timeOfDay.After(t.timeFrom) || timeOfDay.Equal(t.timeFrom) || timeOfDay.Before(t.timeTo)
+func (t relativeTimeSpan) isTimeOfDayInRange(timeOfDay dayTime) bool {
+	if t.timeFrom > t.timeTo { // check if range wraps across days
+		return timeOfDay >= t.timeFrom || timeOfDay < t.timeTo
 	}
 
-	return (t.timeFrom.Before(timeOfDay) || t.timeFrom.Equal(timeOfDay)) && t.timeTo.After(timeOfDay)
+	return t.timeFrom <= timeOfDay && t.timeTo > timeOfDay
```
JTaeuber, line 182: Why are these lesser equals and more equals?

JTaeuber (changes requested):

cmd/kubedownscaler/main.go
```diff
@@ -254,7 +230,7 @@ func scanWorkload(
 		return fmt.Errorf("failed to parse namespace layer from annotations: %w", err)
 	}
 
-	layers := values.Layers{&layerWorkload, &layerNamespace, layerCli, layerEnv}
+	layers := values.Layers{&layerWorkload, &layerNamespace, layerCli, layerEnv, layerDefault}
```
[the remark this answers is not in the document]
jonathan-mayer: yes, the order in which they are evaluated is from index 0 to len-1 and the default layers values should only be used when all other layers fail to specify the value.

internal/pkg/values/timespan.go
jonathan-mayer: since timespans match from their "fromTime" inclusively and to their "toTime" exclusively to allow for configuration like this without any possible conflicts:
DowntimePeriod: Mon-Fri 20-8 utc
UptimePeriod: Mon-Fri 8-12 utc

since otherwise 8:00 would match with both

JTaeuber (approved):
LGTM

Issue #121, opened 2025-02-24
Fix leader election flag never set from args


samuel-esp:
## Issue

This little problem escaped from our sight when merging leader election branch (referring 

```go
flag.BoolVar(
	&c.Once,
	"once",
	false,
	"run scan only once (default: false)",
)
flag.BoolVar(
	&c.Once,
	"leader-election",
	false,
	"enables leader election (default: false)",
)
```

### Problem to solve

Correct the assignment

PR #122, opened 2025-02-24
Fix: correction for leader election flag assignment

samuel-esp:
## Motivation

Bug explained in #121 

## Changes

Made a correction on Leader Election flag assignment

## Tests done

- Unit Tests


samuel-esp:
Re opened with correct branch from my fork

Issue #128, opened 2025-03-01
LeaderElection helm value is not parsed properly


samuel-esp:
## Issue

I just realized this expression is not working as expected

```php
{{/*
If replicaCount is greater than 1 leader election is enabled by default.
*/}}
{{- define "go-kube-downscaler.leaderElection" -}}
{{- if (.Values.replicaCount | int | gt 1) }}
{{- end }}
```

### Problem to solve

Refactor this expression to a more robust expression

PR #67, opened 2024-11-04
Feat/add helm docs

JTaeuber:
## Motivation
Adding documentation for the Helm Chart

Part of #38 

Blocked by #80 and #69

## Changes
* fixed pre-commit email
* refactored values.yaml
* added documentation pages for the helm chart

## Tests done
Ran the site locally

## TODO
Make links to arguments and environment variable pages.

PR #129, opened 2025-03-01
Fix/Leader Election chart bug

samuel-esp:
## Motivation

Solving #128 

## Changes

Refactored `_helpers.tpl` with a new expression for `LeaderElection`

## Tests done

- Unit Tests


JTaeuber:
Funnily enough i also noticed this in https://github.com/caas-team/GoKubeDownscaler/pull/67/commits/540ac55a7d8940c0058a0c4539bf578c6eb02317.

At the same time i noticed that there is an error in the way the role.yaml is executed when the downscaler is constrained.
That should also be fixed with the change implemented in the commit above.

If you want to add it to this pr as well we can fix both here at the same time. 👍🏻

samuel-esp:
Oh I didn't read that branch, I think your commit already covers everything. No point to keep this PR opened

PR #134, opened 2025-03-04
Feat/improve documentation on order of arguments

### LINKED ISSUE #120: Allow timespans in force up-/downtime
## Issue

Currently force uptime and force downtime only support boolean values. This is different to how the pykubedownscaler handled this (see [the Readme](https://github.com/caas-team/py-kube-downscaler/blob/main/README.md?plain=1#L715-L718)). 

### Problem to solve

The problem arises here, since when you want to exclude an absolute span of time while still having the weekly schedule (via uptime) the only way to set this downtime is to migrate your configuration to downtime or downtime period configuration (see [py-kube-downscaler#124](https://github.com/caas-team/py-kube-downscaler/issues/124) for an example where this would be useful)

### Proposal

Change the force up and force downtime to allow for timespan values.
Update [the docs](#38) to include not only documentation on these new timespan values, but also explain the exact order in which force up/downtime downtime/uptime/downtime period/uptime period are taken into account. Also update the docs to include the order of the exclusion values (i think its if its excluded by any of the exclusion configurations). 

With this implemented people will be able to force downscale during specific periods of time without needing to adjust the rest of their configuration.

### Who can address the issue

Anyone with some go experience

jonathan-mayer:
## Motivation

Closes #120 

## Changes

Added docs on boolean timespans, order of values within a layer and layer groups.
Additionally added error message if global references are used to reference the file they are defined in

## TODO

- [x] evaluate renaming layers to scopes


JTaeuber:
Small change, but arguments of funcs are still called l instead of s for scopes

internal/pkg/values/layer.go
JTaeuber, line 1: probably should rename the file as well

internal/pkg/values/layerParser.go
JTaeuber, line 1: same here

internal/pkg/values/layer_test.go
JTaeuber, line 1: and here

website/content/guides/1 - developing/1 - Website.mdx
```diff
@@ -205,7 +205,7 @@ re-save your file to initiate a rebuild or
 
 Assets like PNGs, SVGs, [Excalidraw](#excalidraw) or other files can be useful to have in documentation.
 
-Assets should be placed in a `assets` folder in the folder in which the asset is being used.
+Assets should be placed in an `assets` folder in the directory where it's is being used.
```
JTaeuber, line 208:
```suggestion
Assets should be placed in an `assets` folder in the directory where it is being used.
```

JTaeuber: I think this might have been my mistake 😅

JTaeuber (changes requested):

Issue #111, opened 2025-02-12
Make The User Aware Of How To Use Concurrency


samuel-esp:
## Issue

Actually the default values for requests and limits for the Downscaler are set like this:

```yaml
resources:
  limits:
    cpu: 500m
    memory: 900Mi
  requests:
    cpu: 200m
    memory: 300Mi
```

Using a value of 500m in `resources.limits.cpu` will prevent the user from fully leveraging the concurrency in the code. An ideal value would be at least 2000m (in this way the user would take advantage of at least 2 full vCPUs if possible)

This is what happens for example for IstioProxy 

### Proposal

We could solve this "soft" problem in many possible ways

1) Make the user aware of this inside the docs
2) Set 2000m as the new default
3) Introducing a dedicated a variable in the Helm Chart that the user could leverage to specify its concurrency preference. If the variable is enabled the chart will check if `limit.cpu >= 2000m`, if not it will throw an error

jonathan-mayer:
> 1. Make the user aware of this inside the docs
> 2. Set 2000m as the new default

I would go with one or both of these.

jonathan-mayer:
I would also see this as merged although not in main yet

PR #135, opened 2025-03-04
Docs/Added Parallelism Page

samuel-esp:
## Motivation

To explain users how to leverage parallelism, as stated in #111 

## Changes

- Added doc page

## Tests done

- Unit Tests


samuel-esp:
@JTaeuber @jonathan-mayer we should decide if incrementing CPU Limit to 2000m by default as stated in the comment made by Jonathan in #111

jonathan-mayer (changes requested):
just some tweaks to improve readability in plain text mode.
I am debating though if this should be a guide.

jonathan-mayer:
> @JTaeuber @jonathan-mayer we should decide if incrementing CPU Limit to 2000m by default as stated in the comment made by Jonathan in #111

the only argument against it would be since it could reduce the cpu usage spike and it would only spread out scaling over some time

if we do it i would add a part to this page you added which explains that one can lower the limit and the benefits/ it would have and possible problems it would bring

samuel-esp:
> the only argument against it would be since it could reduce the cpu usage spike and it would only spread out scaling over some time

So you mean to specify:
- CPU Limits help with reducing usage spike
- Underline in the docs that setting CPU Limits doesn't guarantee to be able to always use 2,3,4... full cores

> I am debating though if this should be a guide.

Would you like to put this info elsewhere?

jonathan-mayer:
> > the only argument against it would be since it could reduce the cpu usage spike and it would only spread out scaling over some time
> 
> So you mean to specify:
> 
> * CPU Limits help with reducing usage spike
> * Underline in the docs that setting CPU Limits doesn't guarantee to be able to always use 2,3,4... full cores
> 
> > I am debating though if this should be a guide.
> 
> Would you like to put this info elsewhere?

Yeah, basically just set default limit to 2vcpus, add a part explaining that to reduce spikes you can use lower values and putting this all in the guides instead of the docs

website/content/guides/0 - getting-started/2 - Parallelism.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+You should consider whether the default should be adjusted (higher or lower) based on your cluster size and workload
+to be scaled.
+Most use cases can be handled with a lower CPU Limit value.
+
+When choosing the correct values, the following considerations should be taken into account:
+
+1. Setting a high CPU limit can lead to usage spikes.
+2. The CPU Limit doesn't guarantee that the Downscaler will use the full amount of CPU declared; the pod can expand its
+   CPU usage as long as the resources are available on the node at any given time.
```
jonathan-mayer, line 41:
```suggestion
2. The CPU Limit doesn't guarantee that the Downscaler will use the full amount of CPU declared.
    The pod can expand its CPU usage as long as the resources are available on the node at any given time.
```

website/content/guides/0 - getting-started/2 - Parallelism.mdx
```diff
@@ -0,0 +1,43 @@
...
+  limits:
+    cpu: 2000m
+    memory: 900Mi
+  requests:
+    cpu: 200m
+    memory: 300Mi
+```
+
+You should consider whether the default should be adjusted (higher or lower) based on your cluster size and workload
+to be scaled.
```
jonathan-mayer, line 34:
```suggestion
You should consider whether the default should be adjusted (higher or lower) 
based on your cluster size and workload to be scaled.
```

jonathan-mayer (changes requested):

JTaeuber:
@samuel-esp We should probably also adjust the cpu limit in `website/content/docs/1 - helm-chart/2 - values/11 - resources.mdx`

Issue #137, opened 2025-03-06
refactor to avoid parsing namespace scopes multiple times


jonathan-mayer:
## Issue

Currently the namespace scope is being parsed from the annotations on the namespace every scan for every workload.
This means that in one scan every scan every namespace is gotten from Kubernetes and their annotations parsed for every workload in the namespace.

### Problem to solve

This is a lot of api calls and compute resources we can shave of.

### Proposal

Evaluate if we should:
- Get all namespaces (all namespaces given in the include namespaces runtime configuration) and parse all of them, save the parsed results and access them when needed.
- or we should group the workloads by namespace and only get the namespace once and give the parsed results into a func where the workloads are evaluated.

and implement your chosen solution.

### Who can address the issue

someone with so go knowlege

jonathan-mayer:
you might need to add list namespaces to the helm chart permissions (depending on implementation and if the branch where the helm chart permissions were made more strictly is already merged)

PR #142, opened 2025-03-12
Feat/avoid parsing namespace scope multiple times

samuel-esp:
## Motivation

As stated in #137, this PR aims in reducing API Calls

## Changes

- Introduced `NamespaceToAnnotations` map that stores the annotations for each namespace. In this way namespaces are parsed once before scanning each workload
- The main logic is refactored to do a map lookup instead of an API Call

## Tests done

Unit Tests

## TODO

- [ ] Discussion To Improve The Solution


samuel-esp:
@jonathan-mayer personally I'm looking for a suggestion on how to implement an elegant error handling inside GoFunc. The main problem with my implementation right now is that the GetAnnotations is not error-handled 

Feel free to suggest everything, this is a short MVP i wanted to code quickly

jonathan-mayer (changes requested):
About handling errors in that goroutine. I do see where you are coming from. And we should defnitly handle such an error harshly, basically just like failing to get workloads.
We might just have to make a channel for the errors in an select statement where the other opiton would be the waitgroup finishing

cmd/kubedownscaler/main.go
```diff
@@ -191,19 +196,79 @@ func startScanning(
 	return nil
 }
 
+func GetAnnotationsFromNamespace(
```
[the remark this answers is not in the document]
samuel-esp: for the moment I unexported, I'll leave this open so we can decide after the current refactoring

cmd/kubedownscaler/main.go
```diff
@@ -191,19 +196,79 @@ func startScanning(
...
+			namespace := workload.GetNamespace()
+
+			mutex.Lock()
+			if _, exists := namespaceSet[namespace]; !exists {
+				namespaceSet[namespace] = struct{}{}
+
+				slog.Debug("visited namespace", "namespace", namespace)
+			}
+
+			namespaceToAnnotations[namespace] = make(map[string]string)
```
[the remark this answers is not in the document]
samuel-esp: removed

cmd/kubedownscaler/main.go
```diff
@@ -191,19 +196,79 @@ func startScanning(
...
 
+func GetAnnotationsFromNamespace(
+	workloads []scalable.Workload,
+	client kubernetes.Client,
+	ctx context.Context,
+) map[string]map[string]string {
+	var waitGroup sync.WaitGroup
+	var mutex sync.Mutex
+	namespaceSet := make(map[string]struct{})
+	namespaceToAnnotations := make(map[string]map[string]string)
```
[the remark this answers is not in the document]
samuel-esp: perfect, I refactored in this way. Creating also namespaceLogger to handle the logEvent for the namespace

samuel-esp:
Actually i modified in this way

- The function that computes the namespace layers was moved to the Client. For this i had to create the `namespaceLogger` in order to directly compute the layers without saving the annotations inside a temporary map. (Thanks to Jonathan I noticed that even if the namespace is retrieved once, the layer was still computed multiple times with my mvp implementation)
- Removed the unnecessary go func to parse namespace from workloads

We should decide:
- If keeping that function inside the kubernetes Client
- if keeping `values.NewLayerPtr()` construct to return a LayerPtr (having 2 constructors could be beneficial in case we need to instantiate Layer as a pointer or as a value, without refactoring `values.NewLayer()` every time according to a potential new use case)

internal/api/kubernetes/client.go
```diff
@@ -281,3 +347,66 @@ func (c client) CreateLease(leaseName string) (*resourcelock.LeaseLock, error) {
...
+
+	namespaceLayers := make(map[string]*values.Layer, len(namespaceSet))
+	errChan := make(chan error, len(namespaceSet))
+
+	waitGroup.Wait()
+
+	for namespace := range namespaceSet {
+		waitGroup.Add(1)
+
+		namespaceLayers[namespace] = values.NewLayerPtr()
```
jonathan-mayer, line 374: I actually just realised this will also cause race conditions since some go routines reading from the map are already running until the last one is initialized. can you put this initialization in another for loop before this one

internal/pkg/values/layer.go
```diff
@@ -59,6 +59,14 @@ func NewLayer() Layer {
 	}
 }
 
+// NewLayerPtr gets a new layer with the default values and returns a pointer to it.
+func NewLayerPtr() *Layer {
```
jonathan-mayer, line 63: yea, i dont like having the two even in the slightest. either we change the NewLayer to return a *Layer everywhere or we have to initialize a variable and reference it when we need a *Layer. All in all we should ever only need *Layer since we should never pass Layer around with how big it is

internal/api/kubernetes/client.go
```diff
@@ -257,6 +263,66 @@ func (c client) addWorkloadEvent(eventType, reason, identifier, message string,
...
+
+	// create event
+	_, err := eventsClient.Create(ctx, &corev1.Event{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+		InvolvedObject: corev1.ObjectReference{
+			Kind:       "Namespace",
+			Namespace:  namespace,
```
jonathan-mayer, line 306: namespaces are referenced without the namespace field being set
image

jonathan-mayer (changes requested):
I think we need to refactor the resource/namespace loggers all together. The client addEvent functions can be combined. If possible the types should also both be in one or at least the they should only have the nessessary methods

internal/pkg/values/layer.go
samuel-esp: refactored to only return *Layer

samuel-esp:
Still need to rebase to the latest changes, I'll do it asap

internal/api/kubernetes/client.go
```diff
@@ -281,3 +324,65 @@ func (c client) CreateLease(leaseName string) (*resourcelock.LeaseLock, error) {
...
+			slog.Debug("visited namespace", "namespace", workload.GetNamespace())
+		}
+	}
+
+	namespaceLayers := make(map[string]*values.Scope, len(namespaceSet))
+	errChan := make(chan error, len(namespaceSet))
+
+	for namespace := range namespaceSet {
+		scope := values.NewScope()
+		namespaceLayers[namespace] = scope
```
jonathan-mayer, line 346:
```suggestion
		namespaceLayers[namespace] = values.NewScope()
```

internal/api/kubernetes/client.go
```diff
@@ -281,3 +324,65 @@ func (c client) CreateLease(leaseName string) (*resourcelock.LeaseLock, error) {
...
+	}
+
+	for namespace := range namespaceSet {
+		waitGroup.Add(1)
+
+		go func(namespace string) {
```
jonathan-mayer, line 352: we might want to pass in ctx here too since we dont want all goroutines using the context of the parent function at the same time

internal/api/kubernetes/logger.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+}
+
+// NewNamespaceLogger creates a logger for namespaces.
+func NewNamespaceLogger(client Client, namespace string) Logger {
+	return Logger{
+		client:    client,
+		namespace: namespace,
+	}
+}
```
jonathan-mayer, line 34: I think a better abstraction would be:
```go
type ResourceLogger struct {
	logger resourceLogger
}

func (r ResourceLogger) ErrorInvalidAnnotation(/*...*/) {/*...building on r.logger...*/}
func (r ResourceLogger) ErrorInvalidAnnotation(/*...*/) {/*...building on r.logger...*/}

type resourceLogger interface {
	log(/*all required basic log parameters*/)
}

type namespaceLogger struct {
	client Client
	namespace string
}

func (n namespaceLogger) log(/*...*/) {/*...based on client.addEvent...*/}

type workloadLogger struct {
	client Client
	workload scalable.Workload
}

func (w workloadLogger) log(/*...*/) {/*...based on client.addEvent...*/}
```

cmd/kubedownscaler/main.go
```diff
@@ -258,23 +263,16 @@ func scanWorkload(
 	)
 
 	scopeWorkload := values.NewScope()
-	if err = scopeWorkload.GetScopeFromAnnotations(workload.GetAnnotations(), resourceLogger, ctx); err != nil {
+	if err := scopeWorkload.GetScopeFromAnnotations(workload.GetAnnotations(), resourceLogger, ctx); err != nil {
```
jonathan-mayer, line 266: maybe we should add a `var err error` at the top of this function to avoid allocating the error variable multiple times

jonathan-mayer (approved):

jonathan-mayer:
wasn't supposed to be a approval🫠

samuel-esp:
Should be quick, I'll refactor by launch

jonathan-mayer (approved):

PR #144, opened 2025-03-13
fix: err on missing time annotation

### LINKED ISSUE #141: Enabling --time-annotation/--deployment-time-annotation skip the scaling if timezone/grace-period is not defined
## Issue
When we enable --time-annotation flag. It skips all the deployment scaling.

Enable --time-annotation. Now don't set any annotations it will skip all the deployments with error logging
`failed to scan workload error="failed to get if workload is on grace period: error: annotation isn't set on workload" workload=<workload>  namespace=<namespace>`
### Problem to solve

Ideally enabling any feature should not break anything. What we should have is enable it globally and use default untill and unles time-annotation is defined in workload. 
Or give a way to use annotation directly on the workload instead of global cli config

jonathan-mayer:
## Motivation

closes #141 

## Changes

the downscaler will now default to the creationTime when the time annotation isn't set on a workload instead of erroring and skipping the workload

### DIFF
--- a/internal/pkg/values/scope.go
+++ b/internal/pkg/values/scope.go
@@ -16,7 +16,6 @@ var (
 	errTimeAndPeriod            = errors.New("error: both a time and a period is defined")
 	errInvalidDownscaleReplicas = errors.New("error: downscale replicas value is invalid")
 	errValueNotSet              = errors.New("error: no scope implements this value")
-	errAnnotationNotSet         = errors.New("error: annotation isn't set on workload")
 )
 
 // Scaling is an enum that describes the current Scaling.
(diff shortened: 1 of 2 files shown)


JTaeuber:
I think this looks good but maybe we should merge this after #147 since it changes the layer / scope files again.

Saurav12r:
https://github.com/caas-team/GoKubeDownscaler/pull/147 is already merged. But there are conflicts.
Can this also be merged also after resolving conflict?

jonathan-mayer:
@Saurav12r just resolved the changes. We will have to see if we actually make an release already or wait on some of the other prs/issues.

JTaeuber (approved):
LGTM

PR #159, opened 2025-03-20
chore: release v1.2.0

jonathan-mayer:
## Motivation

Creates release v1.2.0.

Seeing as that we had [46 commits](https://github.com/caas-team/GoKubeDownscaler/compare/v1.1.1...main) to main now including features, fixes and security vulnerabilities i don't see why we cant release where we are at right now.
I think #144 should be included in this so we'll have to wait until it is merged.


jonathan-mayer:
ok, the message should be fixed once the changed workflow is on main.

JTaeuber (approved):
👍🏻

PR #164, opened 2025-03-21
chore: fix package name

JTaeuber:
## Motivation
Fixes the package name so the workflow won't fail anymore. 😅 

## Changes
- chore: change package name in tag action


jonathan-mayer (approved):

jonathan-mayer (approved):
LFTM

jonathan-mayer (approved):

PR #167, opened 2025-03-21
chore: add keep younger than

JTaeuber:
## Motivation
We should keep untagged versions younger than one day to avoid deleting needed packages.

## Changes
- add keep younger than


jonathan-mayer (approved):
LG™

Issue #165, opened 2025-03-21
Distinguish resources with or without ownerReference


samuel-esp:
## Issue

Right now the KubeDownscaler, during the scaling process, cannot distinguish between resources that have ownerReferences or not. For example if you put `jobs` inside `--includeResources` it will scale both "original" jobs and jobs originated from `CronJobs`. This is a potentially unwanted behavior that should be refactored in a way that we are able to distinguish the Job origin 

While `Job` is for sure the resource the most affected resource in our context, this behavior could be extended to other resources in the future and it will be useful to include a logic that evaluates that

**Example**: in the future if we want to support `Pod` resources, we should be able to distinguish mirror pods (standalone pods) from pods orginated by a `Deployment` (which have `ReplicaSet` ownerReference) or by `Job` (which have `Job` as ownerReference). The user may want to scale down only mirror pods

### Proposal

- When scaling, we should check the `ownerReference` field, if the `ownerReference` comes from a resource that should be scaled (because it is inside `--includeResources`) then it will be scaled otherwise it will be skipped

### Who Can Address the Issue

- Anyone with Go knowledge

Issue #157, opened 2025-03-19
Draft: Make ScalableResources Data Driven


jonathan-mayer:
### Problem to Solve

Currently to add new scalable resources we have to do a bunch of copying and the user has no way of simply adding one without going through a pr here.

### Proposal

Make the resources that can be scaled via the gokubedownscaler data driven.
This means:
- [ ] make scaling drivers (sets spec.Replica, sets spec.minReplicas, sets spec.suspend, sets some label etc.)
- [ ] 
- [ ] by doing this we will for sure have to give up not using something like the dynamic client
- [ ] change the permissions in the helm chart (this should still be done dynamically)

- [ ] update the docs to reflect the changes:
  - [ ] add part to explain the way to add new scalable resources

### Who Can Address the Issue

I will try to implement this in the beginning of April

### Other Links/References

https://github.com/caas-team/py-kube-downscaler/issues/145#issuecomment-2736588553

jonathan-mayer:
I forgot if i still wanted to add something here so for now i removed the "Draft: "

PR #131, opened 2025-03-02
Feat/admission controller

samuel-esp:
## Motivation

Adding the Admission Controller feature as mentioned in #32. Differently from py-kube-downscaler. The Admission Controller feature will be implemented with a new custom component that will share the same internal logic of the Downscaler allowing enhanced capabilities when compared to the Admission Controller present in py-kube-downscaler

Sharing the same internal logic will also allow us to implement new admission layers once, speeding up development task (that previously were implemented twice, for Gatekeeper and Kyverno)

Admission Logic Flow:

```mermaid
flowchart LR
    A[New Resource] --> B{Is Resource Included In Downscaler Arguments?}
    B -- No --> C[Create Normally]
    B -- Yes --> D[Intercepted by ValidatingWebhookConfiguration]
    D --> E[Admission Controller]
    E --> F{Should the resource be downscaled right now?}
    F -- ✅ Approve --> G[Resource Created]
    F -- ❌ Deny --> H[Blocked]
```

I'm opening the PR but there is still much to discuss and agree about

## Changes

- Added new component (http web server)
- Refactored argument parser
- Added custom logic for handling admission
- Added Kubernetes Objects for Admission Controller in Helm Chart

## Tests done

- Unit Tests should be added

## TODO

- [ ] Discuss how to refactor `FilterExcluded` function in order to accomodate a single workload
- [ ] Discuss evaluation layers to use when admitting workloads
- [ ] Unit Tests
- [ ] Discuss how to handle tls.crt, tls.key, caBundle in Helm Chart
- [ ] Discuss project structure to accomodate multiple components
- [ ] Refactor CICD Pipeline to build multiple Dockerfile

PR #171, opened 2025-03-25
Feat/scale workload only if its ownerReference is among includedResources

samuel-esp:
## Motivation

As stated in #165 right now the KubeDownscaler cannot distinguish between resources that have ownerReferences or not during the scaling process

## Changes

- Refactored FilterExcluded (specifically the isWorkloadExcluded sub-function) to check also for ownerReference when evaluating workload scaling
- Introduced some other functions to support the logic explained above

## Tests Done

- Unit Tests

## TODO

- [ ] Live Tests

### DIFF
--- a/cmd/kubedownscaler/errors.go
+++ b/cmd/kubedownscaler/errors.go
@@ -14,14 +14,14 @@ func (n *NamespaceScopeRetrieveError) Error() string {
 	return fmt.Sprintf("failed to get namespace scope for namespace %q", n.namespace)
 }
 
-type MaxRetriesExceeded struct {
+type MaxRetriesExceededError struct {
 	maxRetries int
 }
 
 func newMaxRetriesExceeded(maxRetries int) error {
-	return &MaxRetriesExceeded{maxRetries: maxRetries}
+	return &MaxRetriesExceededError{maxRetries: maxRetries}
 }
 
-func (m *MaxRetriesExceeded) Error() string {
+func (m *MaxRetriesExceededError) Error() string {
 	return fmt.Sprintf("failed to scale resource: number of max retries exceeded (%d) will try again in the next cycle", m.maxRetries)
 }
--- a/cmd/kubedownscaler/errors.go
+++ b/cmd/kubedownscaler/errors.go
@@ -13,3 +13,15 @@ func newNamespaceScopeRetrieveError(namespace string) error {
 func (n *NamespaceScopeRetrieveError) Error() string {
 	return fmt.Sprintf("failed to get namespace scope for namespace %q", n.namespace)
 }
+
+type MaxRetriesExceeded struct {
+	maxRetries int
+}
+
+func newMaxRetriesExceeded(maxRetries int) error {
+	return &MaxRetriesExceeded{maxRetries: maxRetries}
+}
+
+func (m *MaxRetriesExceeded) Error() string {
+	return fmt.Sprintf("failed to scale resource: number of max retries exceeded (%d) will try again in the next cycle", m.maxRetries)
+}
--- a/cmd/kubedownscaler/main.go
+++ b/cmd/kubedownscaler/main.go
@@ -188,7 +188,7 @@ func startScanning(
 
 				defer waitGroup.Done()
 
-				err := attemptScan(client, ctx, scopeDefault, scopeCli, scopeEnv, namespaceScopes, config, workload)
+				err := scanWorkload(workload, client, ctx, scopeDefault, scopeCli, scopeEnv, namespaceScopes, config)
 				if err != nil {
 					slog.Error("failed to scan workload", "error", err, "workload", workload.GetName(), "namespace", workload.GetNamespace())
 					return
@@ -213,27 +213,25 @@ func startScanning(
 	return nil
 }
 
-// attemptScan is a wrapper for scanWorkload to handle retries on concurrent writes.
-func attemptScan(
+// attemptScaling handles retries for scaling a workload in case of conflicts.
+func attemptScaling(
 	client kubernetes.Client,
 	ctx context.Context,
-	scopeDefault, scopeCli, scopeEnv *values.Scope,
-	namespaceScopes map[string]*values.Scope,
-	config *util.RuntimeConfiguration,
+	scaling values.Scaling,
 	workload scalable.Workload,
+	scopes values.Scopes,
+	config *util.RuntimeConfiguration,
 ) error {
-	slog.Debug("scanning workload", "workload", workload.GetName(), "namespace", workload.GetNamespace())
-
 	for retry := range config.MaxRetriesOnConflict + 1 {
-		err := scanWorkload(workload, client, ctx, scopeDefault, scopeCli, scopeEnv, namespaceScopes, config)
+		err := scaleWorkload(scaling, workload, scopes, client, ctx)
 		if err != nil {
-			if !(strings.Contains(err.Error(), registry.OptimisticLockErrorMsg)) {
-				return fmt.Errorf("failed to scan workload: %w", err)
+			if !strings.Contains(err.Error(), registry.OptimisticLockErrorMsg) {
+				return fmt.Errorf("failed to scale workload: %w", err)
 			}
 
 			slog.Warn("workload modified, retrying", "attempt", retry+1, "workload", workload.GetName(), "namespace", workload.GetNamespace())
 
-			err := client.RegetWorkload(workload, ctx)
+			err = client.RegetWorkload(workload, ctx)
 			if err != nil {
 				return fmt.Errorf("failed to fetch updated workload: %w", err)
 			}
@@ -241,18 +239,19 @@ func attemptScan(
 			continue
 		}
 
-		slog.Debug("successfully scanned workload", "workload", workload.GetName(), "namespace", workload.GetNamespace())
+		slog.Debug("successfully scaled workload", "workload", workload.GetName(), "namespace", workload.GetNamespace())
 
 		return nil
 	}
 
-	slog.Error("failed to scan workload", "attempts", config.MaxRetriesOnConflict+1)
+	slog.Error("failed to scale workload", "attempts", config.MaxRetriesOnConflict+1)
 
-	return nil
+	return errors.New("exceeded max retries for scaling workload")
 }
 
 // nolint: cyclop // it is a big function and we can refactor it a bit but it should be fine for now
 // scanWorkload runs a scan on the worklod, determining the scaling and scaling the workload.
+// Updated scanWorkload to use the refactored functions.
 func scanWorkload(
 	workload scalable.Workload,
 	client kubernetes.Client,
@@ -262,14 +261,8 @@ func scanWorkload(
 	config *util.RuntimeConfiguration,
 ) error {
 	resourceLogger := kubernetes.NewResourceLoggerForWorkload(client, workload)
-	var err error
 
-	slog.Debug(
-		"parsing workload scope from annotations",
-		"annotations", workload.GetAnnotations(),
-		"name", workload.GetName(),
-		"namespace", workload.GetNamespace(),
-	)
+	var err error
 
 	scopeWorkload := values.NewScope()
 	if err = scopeWorkload.GetScopeFromAnnotations(workload.GetAnnotations(), resourceLogger, ctx); err != nil {
@@ -282,8 +275,6 @@ func scanWorkload(
 	}
 
 	scopes := values.Scopes{scopeWorkload, scopeNamespace, scopeCli, scopeEnv, scopeDefault}
-
-	slog.Debug("finished parsing all scopes", "scopes", scopes, "workload", workload.GetName(), "namespace", workload.GetNamespace())
 
 	isInGracePeriod, err := scopes.IsInGracePeriod(
 		config.TimeAnnotation,
@@ -312,68 +303,40 @@ func scanWorkload(
 		return nil
 	}
 
-	err = scaleWorkload(scaling, workload, scopes, client, ctx)
+	err = attemptScaling(client, ctx, scaling, workload, scopes, config)
 	if err != nil {
 		return fmt.Errorf("failed to scale workload: %w", err)
 	}
 
-	scaleChildren := scopes.GetScaleChildren()
-	if scaleChildren {
-		err = scaleChildrenWorkloads(scaling, workload, scopes, client, ctx)
+	if scopes.GetScaleChildren() {
+		childrenWorkloads, err := client.GetChildrenWorkloads(workload, ctx)
 		if err != nil {
-			return fmt.Errorf("failed to scale children workloads: %w", err)
+			return fmt.Errorf("failed to get children workloads: %w", err)
 		}
+
+		scaleWorkloads(scaling, childrenWorkloads, scopes, client, ctx, config)
 	}
 
 	return nil
 }
 
-// scaleChildrenWorkloads scales the children workloads of the given workload.
-func scaleChildrenWorkloads(
+// scaleWorkloads triggers scaling for a list of workloads without waiting for results.
+func scaleWorkloads(
 	scaling values.Scaling,
-	workload scalable.Workload,
+	workloads []scalable.Workload,
 	scopes values.Scopes,
 	client kubernetes.Client,
 	ctx context.Context,
-) error {
-	slog.Debug("scaling children workloads of workload", "workload", workload.GetName(), "namespace", workload.GetNamespace())
-
-	childrenWorkloads, err := client.GetChildrenWorkloads(workload, ctx)
-	if err != nil {
-		return fmt.Errorf("failed to get children workloads: %w", err)
-	}
-
-	var waitGroup sync.WaitGroup
-	errCh := make(chan error, len(childrenWorkloads))
-	allErrors := make([]error, 0, len(childrenWorkloads))
-
-	for _, childWorkload := range childrenWorkloads {
-		waitGroup.Add(1)
-
-		go func(childWorkload scalable.Workload) {
-			defer waitGroup.Done()
-			slog.Debug("scaling workload", "workload", childWorkload.GetName(), "namespace", childWorkload.GetNamespace())
-
-			err = scaleWorkload(scaling, childWorkload, scopes, client, ctx)
+	config *util.RuntimeConfiguration,
+) {
+	for _, workload := range workloads {
+		go func(workload scalable.Workload) {
+			err := attemptScaling(client, ctx, scaling, workload, scopes, config)
 			if err != nil {
-				if !strings.Contains(err.Error(), registry.OptimisticLockErrorMsg) {
-					errCh <- fmt.Errorf("failed to scale workload %s: %w", childWorkload.GetName(), err)
-				}
+				slog.Error("failed to scale workload", "error", err, "workload", workload.GetName(), "namespace", workload.GetNamespace())
 			}
-		}(childWorkload)
+		}(workload)
 	}
-
-	waitGroup.Wait()
-
-	for err = range errCh {
-		allErrors = append(allErrors, err)
-	}
-
-	if len(allErrors) > 0 {
-		return errors.Join(allErrors...)
-	}
-
-	return nil
 }
 
 // scaleWorkload scales the given workload according to the given wanted scaling state.
--- a/cmd/kubedownscaler/main.go
+++ b/cmd/kubedownscaler/main.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"context"
-	"errors"
 	"flag"
 	"fmt"
 	"log/slog"
--- a/cmd/kubedownscaler/main.go
+++ b/cmd/kubedownscaler/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"flag"
 	"fmt"
 	"log/slog"
@@ -246,6 +247,7 @@ func attemptScan(
 	return nil
 }
 
+// nolint: cyclop // it is a big function and we can refactor it a bit but it should be fine for now
 // scanWorkload runs a scan on the worklod, determining the scaling and scaling the workload.
 func scanWorkload(
 	workload scalable.Workload,
@@ -309,6 +311,60 @@ func scanWorkload(
 	err = scaleWorkload(scaling, workload, scopes, client, ctx)
 	if err != nil {
 		return fmt.Errorf("failed to scale workload: %w", err)
+	}
+
+	scaleChildren := scopes.GetScaleChildren()
+	if scaleChildren {
+		err = scaleChildrenWorkloads(scaling, workload, scopes, client, ctx)
+		if err != nil {
+			return fmt.Errorf("failed to scale children workloads: %w", err)
+		}
+	}
+
+	return nil
+}
+
+// scaleChildrenWorkloads scales the children workloads of the given workload.
+func scaleChildrenWorkloads(
+	scaling values.Scaling,
+	workload scalable.Workload,
+	scopes values.Scopes,
+	client kubernetes.Client,
+	ctx context.Context,
+) error {
+	slog.Debug("scaling children workloads of workload", "workload", workload.GetName(), "namespace", workload.GetNamespace())
+
+	childrenWorkloads, err := client.GetChildrenWorkloads(workload, ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get children workloads: %w", err)
+	}
+
+	var waitGroup sync.WaitGroup
+	errCh := make(chan error, len(childrenWorkloads))
+	allErrors := make([]error, 0, len(childrenWorkloads))
+
+	for _, childWorkload := range childrenWorkloads {
+		waitGroup.Add(1)
+
+		go func(childWorkload scalable.Workload) {
+			defer waitGroup.Done()
+			slog.Debug("scaling workload", "workload", childWorkload.GetName(), "namespace", childWorkload.GetNamespace())
+
+			err = scaleWorkload(scaling, childWorkload, scopes, client, ctx)
+			if err != nil {
+				errCh <- err
+			}
+		}(childWorkload)
+	}
+
+	waitGroup.Wait()
+
+	for err = range errCh {
+		allErrors = append(allErrors, err)
+	}
+
+	if len(allErrors) > 0 {
+		return errors.Join(allErrors...)
 	}
 
 	return nil
--- a/cmd/kubedownscaler/main.go
+++ b/cmd/kubedownscaler/main.go
@@ -246,7 +246,7 @@ func attemptScaling(
 
 	slog.Error("failed to scale workload", "attempts", config.MaxRetriesOnConflict+1)
 
-	return errors.New("exceeded max retries for scaling workload")
+	return fmt.Errorf("exceeded max retries for scaling workload")
 }
 
 // nolint: cyclop // it is a big function and we can refactor it a bit but it should be fine for now
--- a/cmd/kubedownscaler/main.go
+++ b/cmd/kubedownscaler/main.go
@@ -251,7 +251,6 @@ func attemptScaling(
 
 // nolint: cyclop // it is a big function and we can refactor it a bit but it should be fine for now
 // scanWorkload runs a scan on the worklod, determining the scaling and scaling the workload.
-// Updated scanWorkload to use the refactored functions.
 func scanWorkload(
 	workload scalable.Workload,
 	client kubernetes.Client,
@@ -275,6 +274,8 @@ func scanWorkload(
 	}
 
 	scopes := values.Scopes{scopeWorkload, scopeNamespace, scopeCli, scopeEnv, scopeDefault}
+
+	slog.Debug("finished parsing all scopes", "scopes", scopes, "workload", workload.GetName(), "namespace", workload.GetNamespace())
 
 	isInGracePeriod, err := scopes.IsInGracePeriod(
 		config.TimeAnnotation,
@@ -320,7 +321,7 @@ func scanWorkload(
 	return nil
 }
 
-// scaleWorkloads triggers scaling for a list of workloads without waiting for results.
+// scaleWorkloads scales the given workloads to the specified scaling asynchronously.
 func scaleWorkloads(
 	scaling values.Scaling,
 	workloads []scalable.Workload,
--- a/cmd/kubedownscaler/main.go
+++ b/cmd/kubedownscaler/main.go
@@ -162,7 +162,18 @@ func startScanning(
 			return fmt.Errorf("failed to get workloads: %w", err)
 		}
 
-		workloads = scalable.FilterExcluded(workloads, config.IncludeLabels, config.ExcludeNamespaces, config.ExcludeWorkloads)
+		includedResourcesKindSet, err := client.GetKinds(config.IncludeResources)
+		if err != nil {
+			return fmt.Errorf("failed to get kinds for included resources: %w", err)
+		}
+
+		workloads = scalable.FilterExcluded(
+			workloads,
+			config.IncludeLabels,
+			config.ExcludeNamespaces,
+			config.ExcludeWorkloads,
+			includedResourcesKindSet,
+		)
 		slog.Info("scanning over workloads matching filters", "amount", len(workloads))
 
 		namespaceScopes, err := client.GetNamespaceScopes(workloads, ctx)
--- a/cmd/kubedownscaler/main.go
+++ b/cmd/kubedownscaler/main.go
@@ -157,7 +157,7 @@ func startScanning(
 	for {
 		slog.Info("scanning workloads")
 
-		workloads, workloadsUIDToWorkload, err := client.GetWorkloads(config.IncludeNamespaces, config.IncludeResources, ctx)
+		workloads, err := client.GetWorkloads(config.IncludeNamespaces, config.IncludeResources, ctx)
 		if err != nil {
 			return fmt.Errorf("failed to get workloads: %w", err)
 		}
@@ -173,7 +173,6 @@ func startScanning(
 			config.ExcludeNamespaces,
 			config.ExcludeWorkloads,
 			includedResourcesKindSet,
-			workloadsUIDToWorkload,
 		)
 		slog.Info("scanning over workloads matching filters", "amount", len(workloads))
 
--- a/cmd/kubedownscaler/main.go
+++ b/cmd/kubedownscaler/main.go
@@ -162,17 +162,11 @@ func startScanning(
 			return fmt.Errorf("failed to get workloads: %w", err)
 		}
 
-		includedResourcesKindSet, err := client.GetKinds(config.IncludeResources)
-		if err != nil {
-			return fmt.Errorf("failed to get kinds for included resources: %w", err)
-		}
-
 		workloads = scalable.FilterExcluded(
 			workloads,
 			config.IncludeLabels,
 			config.ExcludeNamespaces,
 			config.ExcludeWorkloads,
-			includedResourcesKindSet,
 		)
 		slog.Info("scanning over workloads matching filters", "amount", len(workloads))
 
--- a/cmd/kubedownscaler/main.go
+++ b/cmd/kubedownscaler/main.go
@@ -352,7 +352,9 @@ func scaleChildrenWorkloads(
 
 			err = scaleWorkload(scaling, childWorkload, scopes, client, ctx)
 			if err != nil {
-				errCh <- err
+				if !strings.Contains(err.Error(), registry.OptimisticLockErrorMsg) {
+					errCh <- fmt.Errorf("failed to scale workload %s: %w", childWorkload.GetName(), err)
+				}
 			}
 		}(childWorkload)
 	}
--- a/cmd/kubedownscaler/main.go
+++ b/cmd/kubedownscaler/main.go
@@ -246,7 +246,8 @@ func attemptScaling(
 
 	slog.Error("failed to scale workload", "attempts", config.MaxRetriesOnConflict+1)
 
-	return errors.New("exceeded max retries for scaling workload")
+	return errMaxRetriesExcedeed
+
 }
 
 // nolint: cyclop // it is a big function and we can refactor it a bit but it should be fine for now
--- a/cmd/kubedownscaler/main.go
+++ b/cmd/kubedownscaler/main.go
@@ -246,7 +246,7 @@ func attemptScaling(
 
 	slog.Error("failed to scale workload", "attempts", config.MaxRetriesOnConflict+1)
 
-	return fmt.Errorf("exceeded max retries for scaling workload")
+	return errors.New("exceeded max retries for scaling workload")
 }
 
 // nolint: cyclop // it is a big function and we can refactor it a bit but it should be fine for now
--- a/cmd/kubedownscaler/main.go
+++ b/cmd/kubedownscaler/main.go
@@ -246,7 +246,7 @@ func attemptScaling(
 
 	slog.Error("failed to scale workload", "attempts", config.MaxRetriesOnConflict+1)
 
-	return errMaxRetriesExcedeed
+	return newMaxRetriesExceeded(config.MaxRetriesOnConflict)
 }
 
 // nolint: cyclop // it is a big function and we can refactor it a bit but it should be fine for now
@@ -262,6 +262,12 @@ func scanWorkload(
 	resourceLogger := kubernetes.NewResourceLoggerForWorkload(client, workload)
 
 	var err error
+	slog.Debug(
+		"parsing workload scope from annotations",
+		"annotations", workload.GetAnnotations(),
+		"name", workload.GetName(),
+		"namespace", workload.GetNamespace(),
+	)
 
 	scopeWorkload := values.NewScope()
 	if err = scopeWorkload.GetScopeFromAnnotations(workload.GetAnnotations(), resourceLogger, ctx); err != nil {
--- a/cmd/kubedownscaler/main.go
+++ b/cmd/kubedownscaler/main.go
@@ -157,7 +157,7 @@ func startScanning(
 	for {
 		slog.Info("scanning workloads")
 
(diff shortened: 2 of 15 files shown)


samuel-esp:
Since a resource could have more than one `ownerReference`, we should decide which behavior to adopt in that case. In my implementation the scale happens only if each `ownerReference` is inside `includedResources`

jonathan-mayer:
Im going to list my thoughts on a way to implement this so that is consistent and easier implementable for #157 here:
Non-Admission Controller:
- If object has owner reference with `controlling` set to `true` -> do not scale (Reason: object is controlled by another resource, which means that if modified it could cause the controlling resource and downscaler to scale up/down again and again. If any owner that is to be scaled with the current configuration exists for that resource, that resource is going to be scaled by the parent once the parent is scaled)
- This makes the workload scope always be determined by the parent
Admission Controller:
- handles admission control for every sub resource type of the resources specified in `include-resources`
  - e.g.: Deployments->ReplicaSets,Pods; CronJobs->Jobs,Pods; etc.
- If resource doesn't have ownerReference with controlling:
  - `true`: we recursively go up until we find a resource without a owner reference with `controlling=true`: (This could be adjusted so that if the seen resource types do not fit any of the sub resource type sets specified we stop recursing. This would then require that all intermediate sub resource types ought to be specified and thereby their creation blocked in the scalable workload definition (e.g. Deployments cannot just say that `[Pods]` is the list of sub resource types, since the pods are not direct children of the Deployment workload))
    - If this owner resource has one of the included resource types we continue
    - The workload scope is specified by the owner resource
    - If the owner is scaled down we block the resource
  - `false`: we do not block it
- The admission controller should also probably only ever block the creation of an resource
- This makes the workload scope always be determined by the parent

The above concept seems to me to be the best and most consistent way to handle #131 #157 and #165. Unless there is a flaw in this I think we should see this as the end goal for all 3 of those issues.

One thing we might want to check is if not scaling any resources with an `controlling=true` owner reference could interfere with any CD like Argo, Flux, etc. In which case we should evaluate a way to specify/know if the controlling owner is going to interfere with scaling to decide if we should scale the resource or not.

internal/pkg/scalable/workload.go
```diff
@@ -52,6 +53,32 @@ func GetWorkloads(resource, namespace string, clientsets *Clientsets, ctx contex
...
+
+	obj, exists := resourceKindMap[resource]
+	if !exists {
+		return "", fmt.Errorf("failed to get kind of type %q: %w", resource, errResourceNotSupported)
+	}
+
+	kind := obj.GetObjectKind().GroupVersionKind().GroupKind().String()
+
+	return kind, nil
+}
```
jonathan-mayer, line 80: Could you just hardcode map the aliases to the correct groupkind here. It is not super nice but its going to get refactored out by #157 anyways

internal/api/kubernetes/client.go
```diff
@@ -132,6 +134,22 @@ func (c client) GetWorkloads(namespaces, resourceTypes []string, ctx context.Con
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to get kind for resource type '%s': %w", resource, err)
+		}
+
+		kinds[kind] = struct{}{}
+	}
+
+	return kinds, nil
+}
+
```
jonathan-mayer, line 152: This should probably be in the `scalable` package

internal/pkg/scalable/util.go
```diff
@@ -167,11 +173,17 @@ func isNamespaceExcluded(workload Workload, excludedNamespaces util.RegexList) b
...
+func isWorkloadExcluded(workload Workload, excludedWorkloads util.RegexList, includedResources map[string]struct{}) bool {
 	if excludedWorkloads == nil {
 		return false
 	}
 
+	for _, ownerReference := range workload.GetOwnerReferences() {
+		if _, exists := includedResources[ownerReference.Kind]; !exists {
+			return true
+		}
+	}
```
jonathan-mayer, line 185: adjust for https://github.com/caas-team/GoKubeDownscaler/pull/171#issuecomment-2765221345 to exclude if has any controlling owner

jonathan-mayer (changes requested):

internal/api/kubernetes/client.go
samuel-esp: @jonathan-mayer I put it here because it thought it was easier to handle the error at the cmd level, I think for the same logic of client.GetWorkloads

```go
func startScanning(
	client kubernetes.Client,
	ctx context.Context,
	scopeDefault, scopeCli, scopeEnv *values.Scope,
	config *util.RuntimeConfiguration,
) error {
	slog.Info("started downscaler")

	for {
		slog.Info("scanning workloads")

		workloads, err := client.GetWorkloads(config.IncludeNamespaces, config.IncludeResources, ctx)
		if err != nil {
			return fmt.Errorf("failed to get workloads: %w", err)
		}

		includedResourcesKindSet, err := client.GetKinds(config.IncludeResources)
		if err != nil {
			return fmt.Errorf("failed to get kinds for included resources: %w", err)
		}
                
               ...
```

Instead of handling the conversion inside `FilterExcluded` function or inside `isWorkloadExcluded`. I don't know if this makes sense, otherwise I'll refactor as you said

samuel-esp:
> 
>     If object has owner reference with controlling set to true -> do not scale (Reason: object is controlled by another resource, which means that if modified it could cause the controlling resource and downscaler to scale up/down again and again. If any owner that is to be scaled with the current configuration exists for that resource, that resource is going to be scaled by the parent once the parent is scaled)
>     This makes the workload scope always be determined by the parent

That's a great idea, with this implementation I think we wouldn't need to rely on `Kinds` to make logics
   

>     Admission Controller:
>     handles admission control for every sub resource type of the resources specified in include-resources
>         e.g.: Deployments->ReplicaSets,Pods; CronJobs->Jobs,Pods; etc.
>     If resource doesn't have ownerReference with controlling:
>         true: we recursively go up until we find a resource without a owner reference with controlling=true: (This could be adjusted so that if the seen resource types do not fit any of the sub resource type sets specified we stop recursing. This would then require that all intermediate sub resource types ought to be specified and thereby their creation blocked in the scalable workload definition (e.g. Deployments cannot just say that [Pods] is the list of sub resource types, since the pods are not direct children of the Deployment workload))
>             If this owner resource has one of the included resource types we continue
>             The workload scope is specified by the owner resource
>             If the owner is scaled down we block the resource
>         false: we do not block it
>     The admission controller should also probably only ever block the creation of an resource
>     This makes the workload scope always be determined by the parent

I don't think we need all of this (still reasoning a bit more in depth about it). I'll explain why:
- The Downscaling component works with a "bottom-up" approach: the downscaler can meets both parent resources and child resources but every time a resource is scanned the downscaler doesn't know that. To check if the resource is a child the only way is going up investigating its ownerReference (if `controller=true` is a child resource, otherwise it is a parent resource)
- The Admission Controller component works with a "top-down" approach: there is no way a child resources is admitted in the cluster if its parent is blocked (for example, there is no way you can encounter a pod created from this route `Deployments->ReplicaSets,Pods` if `Deployments` is inside `includeResources`. This piece of code should handle both cases

```go
// isWorkloadExcluded checks if the workloads name is excluded.
func isWorkloadExcluded(workload Workload, excludedWorkloads util.RegexList, includedResources map[string]struct{}) bool {
	if excludedWorkloads == nil {
		return false
	}

	for _, ownerReference := range workload.GetOwnerReferences() {
		if *ownerReference.Controller {
			return true
		}
	}

	return excludedWorkloads.CheckMatchesAny(workload.GetName())
}
```

I might be wrong, so it is useful to know also your point of views @jonathan-mayer @JTaeuber 

> One thing we might want to check is if not scaling any resources with an controlling=true owner reference could interfere with any CD like Argo, Flux, etc. In which case we should evaluate a way to specify/know if the controlling owner is going to interfere with scaling to decide if we should scale the resource or not.

At this point this would be the only concern to verify

jonathan-mayer:
> The Admission Controller component works with a "top-down" approach: there is no way a child resources is admitted in the cluster if its parent is blocked (for example, there is no way you can encounter a pod created from this route Deployments->ReplicaSets,Pods if Deployments is inside includeResources. So the Admission Controller only works with parent resources and therefore only blocks them

So i had another look at our conversation about the admission controllers purpose, apparently i totally forgot that it actually blocks the creation of the resources itself if they would be downscaled. I do think though that that way of doing it is pretty weird. I am just wondering if maybe the way i proposed could actually be a nicer way. That way it would at least not interfere with the top-most(meaning without any controlling owner) objects creation but still stop any actual work from being done when in a downtime period but not yet downscaled. 

As you might have guessed from our conversation we do not actually use this feature of the downscaler in our clusters, so you are more experienced and know what users are expecting when using the admission controller

samuel-esp:
Actually your route is feasible as well, it's just a matter of making a choice, the difference is:

**Hypothesis A (Blocking Pods)**

We will work at pod level (as we are doing right now with the Downscaler component). At creation time, the Admission Controller will intercept the pod creation admission event, from the pod object extracted from the pod creation admission event we will recreate its history up to the original parent (bottom-up approach). If the original parent is included inside `includedResources` we will block the pod creation

**Hypotesis B (Blocking Parent Resources)**


We will work at parent level. At creation time, the Admission Controller will intercept the resource creation admission event (deployment creation admission event, cronjob creation admission event, ecc...), we will extract the object from the creation admission event and if it is included inside `includedResources` the resource will be blocked (top-down approach)

The final result will be pretty much the same

**Differences**

_Error Granularity_: with the first approach, the developer will still be able to deploy every resource even during downtime and will never see any error. Meaning that if you run a CICD pipeline to deploy to the cluster the pipeline will most likely not fail during the `kubectl apply`. In the second implementation the `kubectl apply` will fail and the user will receive an error message saying it is not possible to create that resource because the namespace is downscaled
_Goal Overlapping_: with the first approach the goal of the AdmissionController will be pretty much the same of the Downscaler, probably it will just bypass the grace period but I should test that

The final result will be pretty much the same

---

Just another point about the solution provided above

```go
// isWorkloadExcluded checks if the workloads name is excluded.
func isWorkloadExcluded(workload Workload, excludedWorkloads util.RegexList, includedResources map[string]struct{}) bool {
	if excludedWorkloads == nil {
		return false
	}

	for _, ownerReference := range workload.GetOwnerReferences() {
		if *ownerReference.Controller {
			return true
		}
	}

	return excludedWorkloads.CheckMatchesAny(workload.GetName())
}
```

Found out a small exception that fails with the code shown above: turning off the parent controller won't always mean its child pod will be stopped too (for example when you pause a `CronJob `the `Pod `created from the `CronJob `won't be stopped if it is running until its completion because it is controlled by the `Job` controller). So we should both check the `controller=true` and if the ownerReference is inside includeResources. I'll try to test that

internal/pkg/scalable/workload.go
samuel-esp: done

jonathan-mayer:
What I think about the options:
- Blocking Pods: relatively compute intensive, replaces all but the downscaling when everything is already running of the normal downscaler
- Blocking Parents: in my eyes could cause unexpected results since unless the resource is managed by a controlling parent this resource is just not going to be deployed at all.
- No Admission Controller: like you said this will cause long uptimes on initial creation, which even setting grace period to a very short duration will not completely fix
- Mutating Admission Controller: fixes the problem described in `No Admission Controller`. has some of the mutating admission controller risks/quirks

samuel-esp:
> I actually think terminating the pods would likely harm more than it would do good.

My fear in this case would be handling cronjobs->jobs->pods that last hours. This is basically an antipattern by definition since jobs shouldn't last that much. I saw many people misusing jobs in this way unfortunately

> Yet another option would actually be a mutating admission controller which if the resource should be in downtime scales it down before it ever gets created

This could be a great compromise

jonathan-mayer:
> >I actually think terminating the pods would likely harm more than it would do good.
> 
> My fear in this case would be handling cronjobs->jobs->pods that last hours. This is basically an antipattern by definition since jobs shouldn't last that much. I saw many people misusing jobs in this way unfortunately

Ok lets see if we can resolve this in a nice, reproducible fashion, 

> >Yet another option would actually be a mutating admission controller which if the resource should be in downtime scales it down before it ever gets created
> 
> This could be a great compromise

Another thing that came to mind is how to best handle errors in the controller.
What Im thinking is that:
- error from parsing scaling values on workload: should block creation (this means that this admission controller also acts as an validator for valid config)
- error from other scaling related things(e.g. retrieving namespace annotations, parsing namespace annotations): might be smart to ignore(meaning to admit without patching it) and just add a warning in the response
- error from external packages (e.g. possibly json parsing): might have to be handled on a case by case basis (if we ignore here we should also add some warning to the response)

samuel-esp:
> > > I actually think terminating the pods would likely harm more than it would do good.
> > 
> > 
> > My fear in this case would be handling cronjobs->jobs->pods that last hours. This is basically an antipattern by definition since jobs shouldn't last that much. I saw many people misusing jobs in this way unfortunately
> 
> Ok lets see if we can resolve this in a nice, reproducible fashion,

I'll try to finish my implementation later in the afternoon

> > > Yet another option would actually be a mutating admission controller which if the resource should be in downtime scales it down before it ever gets created
> > 
> > 
> > This could be a great compromise
> 
> Another thing that came to mind is how to best handle errors in the controller. What Im thinking is that:
> 
>     * error from parsing scaling values on workload: should block creation (this means that this admission controller also acts as an validator for valid config)
> 
>     * error from other scaling related things(e.g. retrieving namespace annotations, parsing namespace annotations): might be smart to ignore(meaning to admit without patching it) and just add a warning in the response
> 
>     * error from external packages (e.g. possibly json parsing): might have to be handled on a case by case basis (if we ignore here we should also add some warning to the response)

I'll move these considerations/discussion to the Admission Controller issue so we can talk there

samuel-esp:
As you correctly said, this is the biggest concern @jonathan-mayer. This topic is trickier than we thought 

immagine

When parent resources are owned by GitOps CRDs, this becomes tricky. We have 2 options at this point:

1.  Supporting each GitOps CRDs in the tool, to enable scaling (most likely we would need to suspend them)
2.  Whitelist GitOps CRDs when checking for the OwnerReference (meaning we know that the checks succeds when we encounter these Kinds)

Potential CRDs include: 

- Application (argoproj.io/v1alpha1)
- HelmRelease (helm.toolkit.fluxcd.io/v1beta1)
- Kustomization (kustomize.toolkit.fluxcd.io/v1beta1)

The logic could check 2 things:

- only the direct ownerReference (Job->CronJob). i.e this means that Job will be included in scaling if it comes from a direct resource that is included
- all ownerReferences (Job->CronJob->Argo Application). i.e this means that Job will be included in scaling only if all the ownereReferences are included, i.e if CronJob and Argo Application are included in scaling (implementation 1) or if CronJob is included and ArgoApplication is whitelisted (implementation 2). (i would go with this implementation)

samuel-esp:
Thanks a lot for testing before I managed to set up a small cluster with both, at this point I'll try to refine my implementation. We still have to choose one of these 2 approaches:

- only the first level of ownerReference (Job->CronJob)
- all levels (Pod->Job->CronJob)

At the end, if we are not planning to include pods, there is no resource that comes to my mind with more than a level of ownership (I might be wrong). However, to be precise, we could check all the ownerReferences and if we encounter at least one ownerReference that is included, then we will scale it

samuel-esp:
Still a bit far from perfect, should we handle map errors inside `getNonControllerOwnerReferences` and `getOwnerReferenceParent` or we could be sure the content always exists?

jonathan-mayer:
> Thanks a lot for testing before I managed to set up a small cluster with both, at this point I'll try to refine my implementation. We still have to choose one of these 2 approaches:
> 
> * only the first level of ownerReference (Job->CronJob)
> * all levels (Pod->Job->CronJob)
> 
> At the end, if we are not planning to include pods, there is no resource that comes to my mind with more than a level of ownership (I might be wrong). However, to be precise, we could check all the ownerReferences and if we encounter at least one ownerReference that is included, then we will scale it

Hey so i had a long thought about this. I think there are two options to handle the Jobs created by Cronjobs (or just child resources):

- Just leaving the started jobs running:
  - very simple to implement (just what my fist comment said)
  - long jobs could be unwanted but are "rare" and defaulting to suspending these scheduled jobs could for sure break things in some cases
   - i think this should at least be the default
 - Going top-down, when a cronjob is downscaled downscale all children recursively:
   - rbac in helm chart could have problems here since children might not be in --included-resources
   - it also could be very hard to do a api request to get all children of the resource

In the case where we implement both the way to switch between the modes should be:
- Annotation on workload(scope value) 
- default for this is defined by the type definition

where we either do both, only setting via the type or just annotations with false as default

samuel-esp:
> rbac in helm chart could have problems here since children might not be in --included-resources

That is another point I missed, great catch. Both a bottom-up approach or top-down approach could be problematic for RBAC. A top down approach is even more complicated given the fact I think the parent doesn't have reference to children (we should use a custom query). This would cause slowdown in performance 

At this point I would say we are left with option 1, I propose you this:

1) checking `ownerReference.Controller==true` to exclude all pods that satisfy that expression. In the future if we see this is problematic we could limit this check to a subset of kinds
2)  we could introduce a new argument "--instant-termination" in case the user wants to retain the current behavior (jobs get terminated instantly without waiting for its parent to be turned off)

For option 2: The user should be made aware that if there are both cronjobs and jobs inside includedResources. If the cronjob is excluded and job is not excluded, there will be a circular reference that will always turn off the child resource

jonathan-mayer:
> checking ownerReference.Controller==true and includedResources[ownerReference.Kind]; exists to exclude all pods that satisfy that expression. In the future if we see this is problematic we could limit this check to a subset of kinds

I do prefer this but why still the checking if the owner is in --included-resources? I mean for one do we really want to scale a cronjobs jobs just because cronjobs aren't scaled. Also this would make it so that the scaling parameters come from the job, this is the reason i suggested the top down instead of bottom up approach.

samuel-esp:
> I do prefer this but why still the checking if the owner is in --included-resources? I mean for one do we really want to scale a cronjobs jobs just because cronjobs aren't scaled

Actually you are right to answer this (I don't know if you meant to write "the owner kind is in --included-resources"). 

I think i'm overthinking it, I understood your example and my logic won't work correctly in that case I 100% agree with you. Actually I would say that checking if kind is in includedResources could be an harmful sometimes and therefore should be discarded. 

If there is a CRD out there that for example owns a Deployment or a CronJob. In this case should we exclude the resource by default in any case or we could allow scaling if we don't manage that CRD? (The last option was the goal i had in mind with the check on the kind). If we go for the first option we should log sometimes, maybe at debug level, to let the user know why that workload was excluded

My fear Is implementing this and breaking some behavior people expect right now with the current version

Also, how would you implement a top down approach without touching the RBAC?

jonathan-mayer:
> My fear Is implementing this and breaking some behavior people expect right now with the current version.

> Eventually, in the case of jobs, to force the instant termination when managed by a cron job. Could we introduce an argument that bypass the check controller==true? Maybe it could be another string list of resources. I think this is not the cleanest thing to do

I do still think a bottom up approach just wouldn't be good. You would have to specify and keep in sync the annotations manually.

> Also, how would you implement a top down approach without touching the RBAC?

RBAC could be manageable when we make it so that just every resource we can scale, we have the rbac for (this does kind of take away what we currently do where the downscaler only has the rbac for what he actually needs).

I have kind of found a way which could make it possible to actually go top-down from an api request standpoint. 

In the List options you can specify a label selector. At least deployments and cronjobs actually do set labels on their children that distinguish them. 
Cronjobs set `batch.kubernetes.io/controller-uid` and `controller-uid`.
Deployments set the label specified in spec.selector.matchLabels. 
In theory we can specify these in the scalable resource definition.

To turn this on we should stick to what i said previously.
> In the case where we implement both the way to switch between the modes should be:
> - Annotation on workload(scope value)
> - default for this is defined by the type definition
>
> where we either do both, only setting via the type or just annotations with false as default

samuel-esp:
Now I get it, I didn't think of that. Great idea!

> In the case where we implement both the way to switch between the modes should be:
> 
>     * Annotation on workload(scope value)
> 
>     * default for this is defined by the type definition
> 
> 
> where we either do both, only setting via the type or just annotations with false as default

Could you do a practical example of them when you have time? Is it a check we can perform inside the `FilterExcluded` function or later on in the flow?

jonathan-mayer:
Hey, sorry for the late reply.

> > In the case where we implement both the way to switch between the modes should be:
> > ```
> > * Annotation on workload(scope value)
> > 
> > * default for this is defined by the type definition
> > ```
> > 
> > where we either do both, only setting via the type or just annotations with false as default
> 
> Could you do a practical example of them when you have time? Is it a check we can perform inside the `FilterExcluded` function or later on in the flow?

It would basically already be like this: 
- Filter out any workload that has controlling owner
If the value is set for scaling children on a workload we will:
- "look" at the resource definition(in non-data driven implementation this would likely mean a method like GetChildren on the workload) for how to get its children, and get its children of the resource kinds we know how to downscale
- downscale the children based on the layers of the parent

internal/api/kubernetes/client.go
```diff
@@ -26,6 +27,8 @@ const (
 )
 
 // Client is an interface representing a high-level client to get and modify Kubernetes resources.
+//
+
```
jonathan-mayer, line 31:
```suggestion
```

internal/api/kubernetes/client.go
```diff
@@ -129,6 +136,12 @@ func (c client) GetWorkloads(namespaces, resourceTypes []string, ctx context.Con
 		}
 	}
 
+	UIDToWorkloadMap := make(map[types.UID]*scalable.Workload, len(results))
+
+	for i := range results {
+		UIDToWorkloadMap[results[i].GetUID()] = &results[i] // Store pointer to the workload
+	}
+
```
jonathan-mayer, line 144:
```suggestion
```

internal/pkg/scalable/util.go
```diff
@@ -167,14 +172,34 @@ func isNamespaceExcluded(workload Workload, excludedNamespaces util.RegexList) b
...
+
+	for _, ownerReference := range workload.GetOwnerReferences() {
+		if *ownerReference.Controller {
+			isExcluded = true
+		}
+	}
+
+	return isExcluded
+}
+
```
jonathan-mayer, line 202:
```suggestion
// isManagedByOwnerReference checks if the workload is managed by an controlling owner reference
func isManagedByOwnerReference(workload Workload) bool {
	for _, ownerReference := range workload.GetOwnerReferences() {
		if *ownerReference.Controller {
			return true
		}
	}

	return false
}

```

internal/pkg/scalable/util_test.go
```diff
@@ -91,13 +91,15 @@ func TestFilterExcluded(t *testing.T) {
...
 	}
+
 	tests := []struct {
 		name               string
 		workloads          []Workload
 		includeLabels      util.RegexList
 		excludedNamespaces util.RegexList
 		excludedWorkloads  util.RegexList
 		want               []Workload
+		includedResources  map[string]struct{}
```
jonathan-mayer, line 102: This field doesnt get used anywhere

jonathan-mayer (changes requested):

internal/pkg/scalable/util_test.go
samuel-esp: forgot to delete that in the refactoring

internal/api/kubernetes/client.go
```diff
@@ -26,6 +26,7 @@ const (
 )
 
 // Client is an interface representing a high-level client to get and modify Kubernetes resources.
+
 type Client interface {
 	// GetNamespaceAnnotations gets the annotations of the workload's namespace
```
jonathan-mayer, line 31:
```suggestion
// Client is an interface representing a high-level client to get and modify Kubernetes resources.
type Client interface {
	// GetNamespaceAnnotations gets the annotations of the workload's namespace
```
I think my prior suggestion for some reason didn't delete that line

jonathan-mayer (changes requested):

samuel-esp:
Hey @jonathan-mayer I tried to implement an MVP. Not a great fan of my own implementation, definitely some bloated code could be improved. 

Unfortunately it is a bit different from the talk we had the other day: from what I saw the `controller-uid` annotation inside each Job is not set with the CronJob uid, rather it seems it is a random uid created by the Job itself (see the yaml files below). Anyway that's not a problem because I was able to get the childs name directly from the `.status.active` of the CronJob

I hope the data driven refactoring won't be complicated because of this, anyway we still can ditch this without problems or even limiting the resources where this feature is applied

One question: should we handle the cases with optimistic lock errors (e.g implementing retries also in the scalingChildrenWorkloads function)?

```yaml
apiVersion: v1
items:
- apiVersion: batch/v1
  kind: CronJob
  metadata:
    creationTimestamp: "2024-12-21T17:35:52Z"
    generation: 13
    name: test-cj
    namespace: orange
    resourceVersion: "225361"
    uid: 4a6e4a30-c474-4bc5-9bf5-47f29430fb41
  spec:
    concurrencyPolicy: Allow
    failedJobsHistoryLimit: 1
    jobTemplate:
      metadata:
        creationTimestamp: null
        name: test-cj
      spec:
        template:
          metadata:
            creationTimestamp: null
          spec:
            containers:
            - image: nginx
              imagePullPolicy: Always
              name: test-cj
              resources: {}
              terminationMessagePath: /dev/termination-log
              terminationMessagePolicy: File
            dnsPolicy: ClusterFirst
            restartPolicy: OnFailure
            schedulerName: default-scheduler
            securityContext: {}
            terminationGracePeriodSeconds: 30
    schedule: 0/5 * * * ?
    successfulJobsHistoryLimit: 3
    suspend: false
  status:
    active:
    - apiVersion: batch/v1
      kind: Job
      name: test-cj-29074205
      namespace: orange
      resourceVersion: "224338"
      uid: 0f8e0e24-648b-4472-a1ef-f2152784fa57
    - apiVersion: batch/v1
      kind: Job
      name: test-cj-29074210
      namespace: orange
      resourceVersion: "224598"
      uid: a97d8b02-b4fa-43c1-838d-3adbbbc71413
---
apiVersion: batch/v1
kind: Job
metadata:
  annotations:
    batch.kubernetes.io/cronjob-scheduled-timestamp: "2025-04-12T10:10:00Z"
  creationTimestamp: "2025-04-12T10:10:00Z"
  generation: 1
  labels:
    batch.kubernetes.io/controller-uid: a97d8b02-b4fa-43c1-838d-3adbbbc71413
    batch.kubernetes.io/job-name: test-cj-29074210
    controller-uid: a97d8b02-b4fa-43c1-838d-3adbbbc71413
    job-name: test-cj-29074210
  name: test-cj-29074210
  namespace: orange
  ownerReferences:
  - apiVersion: batch/v1
    blockOwnerDeletion: true
    controller: true
    kind: CronJob
    name: test-cj
    uid: 4a6e4a30-c474-4bc5-9bf5-47f29430fb41
  resourceVersion: "224617"
  uid: a97d8b02-b4fa-43c1-838d-3adbbbc71413
spec:
  backoffLimit: 6
  completionMode: NonIndexed
  completions: 1
  manualSelector: false
  parallelism: 1
  podReplacementPolicy: TerminatingOrFailed
  selector:
    matchLabels:
      batch.kubernetes.io/controller-uid: a97d8b02-b4fa-43c1-838d-3adbbbc71413
  suspend: false
  template:
    metadata:
      creationTimestamp: null
      labels:
        batch.kubernetes.io/controller-uid: a97d8b02-b4fa-43c1-838d-3adbbbc71413
        batch.kubernetes.io/job-name: test-cj-29074210
        controller-uid: a97d8b02-b4fa-43c1-838d-3adbbbc71413
        job-name: test-cj-29074210
    spec:
      containers:
      - image: nginx
        imagePullPolicy: Always
        name: test-cj
        resources: {}
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
      dnsPolicy: ClusterFirst
      restartPolicy: OnFailure
      schedulerName: default-scheduler
      securityContext: {}
      terminationGracePeriodSeconds: 30
status:
  active: 1
  ready: 1
  startTime: "2025-04-12T10:10:00Z"
  terminating: 0
  uncountedTerminatedPods: {}
---
apiVersion: batch/v1
kind: Job
metadata:
  annotations:
    batch.kubernetes.io/cronjob-scheduled-timestamp: "2025-04-12T10:05:00Z"
  creationTimestamp: "2025-04-12T10:08:12Z"
  generation: 1
  labels:
    batch.kubernetes.io/controller-uid: 0f8e0e24-648b-4472-a1ef-f2152784fa57
    batch.kubernetes.io/job-name: test-cj-29074205
    controller-uid: 0f8e0e24-648b-4472-a1ef-f2152784fa57
    job-name: test-cj-29074205
  name: test-cj-29074205
  namespace: orange
  ownerReferences:
  - apiVersion: batch/v1
    blockOwnerDeletion: true
    controller: true
    kind: CronJob
    name: test-cj
    uid: 4a6e4a30-c474-4bc5-9bf5-47f29430fb41
  resourceVersion: "224521"
  uid: 0f8e0e24-648b-4472-a1ef-f2152784fa57
spec:
  backoffLimit: 6
  completionMode: NonIndexed
  completions: 1
  manualSelector: false
  parallelism: 1
  podReplacementPolicy: TerminatingOrFailed
  selector:
    matchLabels:
      batch.kubernetes.io/controller-uid: 0f8e0e24-648b-4472-a1ef-f2152784fa57
  suspend: false
  template:
    metadata:
      creationTimestamp: null
      labels:
        batch.kubernetes.io/controller-uid: 0f8e0e24-648b-4472-a1ef-f2152784fa57
        batch.kubernetes.io/job-name: test-cj-29074205
        controller-uid: 0f8e0e24-648b-4472-a1ef-f2152784fa57
        job-name: test-cj-29074205
    spec:
      containers:
      - image: nginx
        imagePullPolicy: Always
        name: test-cj
        resources: {}
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
      dnsPolicy: ClusterFirst
      restartPolicy: OnFailure
      schedulerName: default-scheduler
      securityContext: {}
      terminationGracePeriodSeconds: 30
status:
  active: 1
  ready: 1
  startTime: "2025-04-12T10:08:12Z"
  terminating: 0
  uncountedTerminatedPods: {}
```

jonathan-mayer:
> One question: should we handle the cases with optimistic lock errors (e.g implementing retries also in the scalingChildrenWorkloads function)?



Have not looked at the code yet but couldnt we just call one of the scanWorkload steps just on the child resource

internal/api/kubernetes/client.go
```diff
@@ -132,6 +138,30 @@ func (c client) GetWorkloads(namespaces, resourceTypes []string, ctx context.Con
...
+		workloads, err := scalable.GetChildrenWorkloads(strings.ToLower(workload.GroupVersionKind().Kind), workload, c.clientsets, ctx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get workloads: %w", err)
+		}
+
+		return append(results, workloads...), nil
+	}
+
+	return []scalable.Workload{}, nil
+}
```
jonathan-mayer, line 163: I think it would be smart to refactor this and scalable.GetChildrenWorkloads like this:

Make a new interface in scalable: 

```go
type ParentWorkload interface {
    GetChildren() ([]scalable.Workload, error)
}
```

and then just check if the workload implements this interface. If it does we get the children and scale them (preferably using a already existing scaleWorkloads function. Although we might have to refactor our existing one to be more generic and have a parent function for getting the layers which will already be provided for the children)

internal/pkg/scalable/util.go
```diff
@@ -167,14 +172,32 @@ func isNamespaceExcluded(workload Workload, excludedNamespaces util.RegexList) b
...
+		return true
+	}
+
 	return excludedWorkloads.CheckMatchesAny(workload.GetName())
 }
 
+// isManagedByOwnerReference checks if the workload is managed by an owner reference that is in the includedResources list.
+func isManagedByOwnerReference(workload Workload) bool {
+	for _, ownerReference := range workload.GetOwnerReferences() {
+		if *ownerReference.Controller {
```
jonathan-mayer, line 193: we should do a nil check here

internal/pkg/values/scope.go
```diff
@@ -55,6 +55,7 @@ func NewScope() Scope {
 	return Scope{
 		DownscaleReplicas: util.Undefined,
 		GracePeriod:       util.Undefined,
+		ScaleChildren:     false,
```
jonathan-mayer, line 58: we wouldnt need to set this here since false is already the zero value

internal/pkg/scalable/cronjobs.go
```diff
@@ -3,17 +3,20 @@ package scalable
...
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // getCronJobs is the getResourceFunc for CronJobs.
 func getCronJobs(namespace string, clientsets *Clientsets, ctx context.Context) ([]Workload, error) {
 	cronjobs, err := clientsets.Kubernetes.BatchV1().CronJobs(namespace).List(ctx, metav1.ListOptions{})
 	if err != nil {
-		return nil, fmt.Errorf("failed to get cronjobs: %w", err)
+		return nil, fmt.Errorf("%w: failed to convert workload to cronjobs", errConversionFailed)
```
jonathan-mayer, line 19: why do we use a custom error here if we get one back

internal/pkg/scalable/cronjobs.go
```diff
@@ -24,6 +27,51 @@ func getCronJobs(namespace string, clientsets *Clientsets, ctx context.Context)
 	return results, nil
 }
 
+// getCronJobsChildren is the getResourceFunc for CronJobs children (Jobs).
+func getCronJobsChildren(workload Workload, clientsets *Clientsets, ctx context.Context) ([]Workload, error) {
+	cronjob, ok := workload.(*suspendScaledWorkload).suspendScaledResource.(*cronJob)
+	if !ok {
+		return nil, fmt.Errorf("%w: %v", errConversionFailed, workload)
+	}
```
jonathan-mayer, line 35: with my suggested refactoring using the `ParentWorkload` interface this shouldn't be nessessary since we would just have a method on `cronJob` here so we can directly use the `c.CronJob` field

internal/pkg/scalable/cronjobs.go
```diff
@@ -24,6 +27,51 @@ func getCronJobs(namespace string, clientsets *Clientsets, ctx context.Context)
...
+		go func(activeJob v1.ObjectReference) {
+			defer waitGroup.Done()
+
+			singleJob, err := clientsets.Kubernetes.BatchV1().Jobs(workload.GetNamespace()).Get(ctx, activeJob.Name, metav1.GetOptions{})
+			if err != nil {
+				errChannel <- fmt.Errorf("failed to get job %s: %w", activeJob.Name, err)
+				return
+			}
+
+			results = append(results, &suspendScaledWorkload{&job{singleJob}})
```
jonathan-mayer, line 56: this could cause a race condition

internal/pkg/scalable/cronjobs.go
```diff
@@ -24,6 +27,51 @@ func getCronJobs(namespace string, clientsets *Clientsets, ctx context.Context)
...
+func getCronJobsChildren(workload Workload, clientsets *Clientsets, ctx context.Context) ([]Workload, error) {
+	cronjob, ok := workload.(*suspendScaledWorkload).suspendScaledResource.(*cronJob)
+	if !ok {
+		return nil, fmt.Errorf("%w: %v", errConversionFailed, workload)
+	}
+
+	activeJobs := cronjob.Status.Active
+
+	var waitGroup sync.WaitGroup
+	errChannel := make(chan error, len(activeJobs))
```
jonathan-mayer, line 40:
```suggestion
	errChannel := make(chan error, 0, len(activeJobs))
```

cmd/kubedownscaler/main.go
jonathan-mayer, line 1: I would refactor it like this:

the conflict handling in attempt scan should only wrap the scaleWorkload function not the scanWorkload function (and renamed accordingly).

we also dont need the results of scaleChildrenWorkloads so we dont need the waitgroup and errCh etc., we can just start the goroutines and let them do their thing.

the scaleChildrenWorkloads could also be renamed to `scaleWorkloads` since it does nothing more than that, trigger scaleWorkload on the array of given workloads

internal/pkg/scalable/cronjobs.go
samuel-esp: this shouldn't be an error i think (?)

jonathan-mayer: my bad 🙃

internal/pkg/values/scope.go
```diff
@@ -225,7 +224,7 @@ func (s Scopes) GetDownscaleReplicas() (int32, error) {
 // GetScaleChildren gets the scale children of the first scope that implements scale children.
 func (s Scopes) GetScaleChildren() bool {
 	for _, scope := range s {
-		if scope.ScaleChildren {
+		if scope.ScaleChildren.isSet && scope.ScaleChildren.value {
 			return true
 		}
```
jonathan-mayer, line 229: It would actually be this, otherwise the tristatebool wouldn't make a difference here

```suggestion
		if scope.ScaleChildren.isSet {
			return scope.ScaleChildren.value
		}
```

internal/pkg/values/scopeParser.go
```diff
@@ -71,7 +71,7 @@ func (s *Scope) ParseScopeFlags() {
 		"the grace period between creation of workload until first downscale (default: 15min)",
 	)
 	flag.BoolVar(
-		&s.ScaleChildren,
+		&s.ScaleChildren.value,
 		"scale-children",
 		false,
```
jonathan-mayer, line 76:
```suggestion
	flag.Var(
		&s.ScaleChildren,
		"scale-children",
```

jonathan-mayer (changes requested):
pre commit also reports some unused dupl nolint directives and that `type getChildResourceFunc is unused`

cmd/kubedownscaler/main.go
```diff
@@ -285,8 +281,6 @@ func scanWorkload(
 
 	scopes := values.Scopes{scopeWorkload, scopeNamespace, scopeCli, scopeEnv, scopeDefault}
 
-	slog.Debug("finished parsing all scopes", "scopes", scopes, "workload", workload.GetName(), "namespace", workload.GetNamespace())
-
```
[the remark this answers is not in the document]
samuel-esp: I worked with my ide and Gedit at the same time, my bad I lost them in the process

cmd/kubedownscaler/main.go
```diff
@@ -26,7 +26,10 @@ const (
 	leaseName = "downscaler-lease"
 )
 
-var errNamespaceLayersRetrieveFailed = errors.New("failed to get namespace layer")
+var (
+	errNamespaceLayersRetrieveFailed = errors.New("failed to get namespace layer")
+	errMaxRetriesExcedeed            = errors.New("exceeded max retries for scaling workload")
+)
```
jonathan-mayer, line 32: please rebase and adjust for custom error handling

jonathan-mayer (changes requested):

cmd/kubedownscaler/main.go
```diff
@@ -254,14 +265,8 @@ func scanWorkload(
...
 ) error {
 	resourceLogger := kubernetes.NewResourceLoggerForWorkload(client, workload)
-	var err error
 
-	slog.Debug(
-		"parsing workload scope from annotations",
-		"annotations", workload.GetAnnotations(),
-		"name", workload.GetName(),
-		"namespace", workload.GetNamespace(),
-	)
```
jonathan-mayer, line 264: This debug log is still missing

cmd/kubedownscaler/main.go
```diff
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"errors"
```
JTaeuber, line 5:
```suggestion
```

cmd/kubedownscaler/errors.go
```diff
@@ -13,3 +13,15 @@ func newNamespaceScopeRetrieveError(namespace string) error {
...
+	maxRetries int
+}
+
+func newMaxRetriesExceeded(maxRetries int) error {
+	return &MaxRetriesExceeded{maxRetries: maxRetries}
+}
+
+func (m *MaxRetriesExceeded) Error() string {
+	return fmt.Sprintf("failed to scale resource: number of max retries exceeded (%d) will try again in the next cycle", m.maxRetries)
+}
```
JTaeuber, line 27:
```suggestion
type MaxRetriesExceededError struct {
	maxRetries int
}

func newMaxRetriesExceededError(maxRetries int) error {
	return &MaxRetriesExceededError{maxRetries: maxRetries}
}

func (m *MaxRetriesExceededError) Error() string {
	return fmt.Sprintf("failed to scale resource: number of max retries exceeded (%d), will try again in the next cycle", m.maxRetries)
}
```

JTaeuber (changes requested):

JTaeuber:
These changes should fix the pre-commit errors.

cmd/kubedownscaler/errors.go
JTaeuber, line 27:
```suggestion
func (n *NamespaceScopeRetrieveError) Error() string {
	return fmt.Sprintf("error: failed to get namespace scope for namespace %q", n.namespace)
}

type MaxRetriesExceededError struct {
	maxRetries int
}

func newMaxRetriesExceededError(maxRetries int) error {
	return &MaxRetriesExceededError{maxRetries: maxRetries}
}

func (m *MaxRetriesExceededError) Error() string {
	return fmt.Sprintf("error: number of max retries exceeded (%d), will try again in the next cycle", m.maxRetries)
}
```

JTaeuber:
These changes should fix the pre-commit errors.

jonathan-mayer (dismissed):

samuel-esp:
After testing I realized I inverted this core logic inside `isWorkloadExcluded`. We should check first if the workload is managed by an owner

```go
if isManagedByOwnerReference(workload) {
  return true
}

if excludedWorkloads == nil {
  return false
}
```

The PR should now be ok, even though I will try to add a test for this specific case

jonathan-mayer (approved):

Issue #175, opened 2025-03-29
ARM64 Builds Please


cablespaghetti:
## Issue

I'd love to deploy this on my AWS Graviton EKS cluster. Pushing a multi-arch image manifest with arm64 images would be amazing.

### Problem to Solve

Running on clusters with arm64-based nodes.

### Further Details

N/A

### Proposal

A change to GitHub Actions configuration.

### Who Can Address the Issue

Not particularly.

JTaeuber:
@cablespaghetti arm64 build should now be available in v1.2.1.

cablespaghetti:
Amazing thank you!

cablespaghetti:
> [@cablespaghetti](https://github.com/cablespaghetti) arm64 build should now be available in v1.2.1.

Just so you know, this has deployed successfully in my cluster. 😄

PR #176, opened 2025-03-30
Feat/multi-arch build

samuel-esp:
## Motivation

As stated in #175, we currently don't build an arm64 image inside the build-image workflow

## Changes

- Added `linux/arm64` inside the supported platforms of `docker_build.yaml`

## Tests Done

- Unit Tests

### DIFF
--- a/.github/workflows/docker_build.yaml
+++ b/.github/workflows/docker_build.yaml
@@ -58,7 +58,7 @@ jobs:
         id: build-push
         with:
           context: .
-          platforms: linux/amd64
+          platforms: linux/amd64,linux/arm64
           push: true
           tags: ${{ steps.meta.outputs.tags }}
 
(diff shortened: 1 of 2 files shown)


jonathan-mayer (dismissed):

jonathan-mayer:
@JTaeuber can we just rerun the workflow for the latest tag so that this is available right away? or if that's not feasible maybe we should consider releasing a new version. Otherwise it might take a few weeks until a new version is going to be released and the arm build is usable

JTaeuber (dismissed):
I think this is fine. 👍🏻 
Maybe just bump the version numbers here, so we can make a release directly?

samuel-esp:
`1.2.1` is it ok?

JTaeuber:
Jup 👍🏻

JTaeuber (approved):

PR #177, opened 2025-03-30
Fix/pre commit not running on forks and bump golangci lint config

jonathan-mayer:
## Motivation

Golangci lint has had a major update which makes the old config invalid. I updated the config using https://golangci-lint.run/product/migration-guide/, this also made the linters complain about some things which i then also adapted.

Ive also noticed that after chaning the pre commit workflow to only run `on_push` made the workflow stop running on forks. (It never ran on forks, it always only ran because the fork had an open pull request)

### DIFF
--- a/.github/workflows/pre_commit.yaml
+++ b/.github/workflows/pre_commit.yaml
@@ -2,6 +2,8 @@ name: Run pre-commit checks
 
 on:
   push:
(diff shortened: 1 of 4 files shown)


JTaeuber (approved):
👍🏻

Issue #7, opened 2024-08-01
Draft: Improve error handling


jonathan-mayer:
## Issue

Currently errors are handled by pre defined errors.New vars. This means that the value that caused the issue can't be shown to the user.

### Problem to solve

There should be custom errors for those cases. They should also have custom values you can set to show the user the value / state that caused this issue.

### Proposal

Implement some custom errors and replace the error vars with them.

### Who can address the issue

Someone who has some experience in Go

PR #198, opened 2025-04-22
Feat/improve error handling

JTaeuber:
## Motivation
Closes https://github.com/caas-team/GoKubeDownscaler/issues/7.

## Changes
- added custom error functions

### DIFF
--- a/internal/pkg/values/dayTime.go
+++ b/internal/pkg/values/dayTime.go
@@ -1,7 +1,6 @@
 package values
 
 import (
-	"errors"
 	"fmt"
 	"strconv"
 	"strings"
@@ -12,8 +11,6 @@ const (
 	Hour   dayTime = 60
 	Minute dayTime = 1
 )
-
-var errTimeOfDayOutOfRange = errors.New("error: the specified time of day is out of range")
 
 func extractDayTime(t time.Time) dayTime {
 	return dayTime(t.Hour())*Hour +
@@ -30,7 +27,7 @@ func parseDayTime(s string) (*dayTime, error) {
 	}
 
 	if hour < 0 || hour > 24 {
-		return nil, errTimeOfDayOutOfRange
+		return nil, newInvalidConfigError("time of day")
 	}
 
 	result += dayTime(hour) * Hour
@@ -41,7 +38,7 @@ func parseDayTime(s string) (*dayTime, error) {
 	}
 
 	if minute < 0 || minute >= 60 {
-		return nil, errTimeOfDayOutOfRange
+		return nil, newInvalidConfigError("time of day")
 	}
 
 	result += dayTime(minute) * Minute
--- a/internal/pkg/values/dayTime.go
+++ b/internal/pkg/values/dayTime.go
@@ -27,7 +27,7 @@ func parseDayTime(s string) (*dayTime, error) {
 	}
 
 	if hour < 0 || hour > 24 {
-		return nil, newInvalidSyntaxError("time of day", "hours must be in a range from 0 to 24")
+		return nil, newInvalidValueError("hours in time of day must be in the range from 0 to 24", strconv.Itoa(hour))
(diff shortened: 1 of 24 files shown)


jonathan-mayer (changes requested):
please unexport all fields in the error structs and the method receivers currently do not all match the structs name

jonathan-mayer (changes requested):

internal/pkg/values/errors.go
```diff
@@ -29,14 +29,15 @@ func (v *ValueNotSetError) Error() string {
...
-func newInvalidConfigError(field string) error {
-	return &InvalidConfigError{field: field}
+func newInvalidSyntaxError(field, reason string) error {
+	return &InvalidSyntaxError{field: field, reason: reason}
 }
 
-func (i *InvalidConfigError) Error() string {
-	return fmt.Sprintf("error: specified %q is invalid", i.field)
+func (i *InvalidSyntaxError) Error() string {
+	return fmt.Sprintf("error: %q is invalid. %q.", i.field, i.reason)
```
jonathan-mayer, line 42: i think in this case this would be more fitting:
```suggestion
	return fmt.Sprintf("error: %q is invalid: %s.", i.field, i.reason)
```

internal/pkg/values/dayTime.go
```diff
@@ -27,7 +27,7 @@ func parseDayTime(s string) (*dayTime, error) {
 	}
 
 	if hour < 0 || hour > 24 {
-		return nil, newInvalidConfigError("time of day")
+		return nil, newInvalidSyntaxError("time of day", "hours must be in a range from 0 to 24")
```
jonathan-mayer, line 30:
```suggestion
		return nil, newInvalidSyntaxError("time of day", "hours must be in the range from 0 to 24")
```

internal/pkg/values/dayTime.go
```diff
@@ -38,7 +38,7 @@ func parseDayTime(s string) (*dayTime, error) {
 	}
 
 	if minute < 0 || minute >= 60 {
-		return nil, newInvalidConfigError("time of day")
+		return nil, newInvalidSyntaxError("time of day", "minutes must be in a range from 0 to 60")
```
jonathan-mayer, line 41:
```suggestion
		return nil, newInvalidSyntaxError("time of day", "minutes must be in the range from 0 to 60")
```

also maybe it would be nice to have this error message "error: "time of day" is invalid: minutes must be in the range from 0 to 60, got 'xx'"

internal/pkg/values/timespan.go
```diff
@@ -110,17 +110,17 @@ func parseRelativeTimeSpan(timespanString string) (*relativeTimeSpan, error) {
...
 	weekdaySpan := strings.Split(parts[0], "-")
 	if len(weekdaySpan) != 2 {
-		return nil, newInvalidConfigError("relative timespan")
+		return nil, newInvalidSyntaxError("relative timespan", "relative timespan has more defined weekdays than expected")
 	}
 
 	timeSpan := strings.Split(parts[1], "-")
 	if len(timeSpan) != 2 {
-		return nil, newInvalidConfigError("relative timespan")
+		return nil, newInvalidSyntaxError("relative timespan", "relative timespan has more defined timespans than expected")
```
jonathan-mayer, line 123: what i just realized is that this could just as well be less of everything so we should change up the messages to reflect on that. Also since its not really "timespans" we might just change all of these up to be more like "the relative timespans weekday span is not in expected format (e.g. "Mon-Fri"), got %q" or
"the relative timespan does not have expected format (e.g. "Mon-Fri 08:00-20:00 UTC"), got %q"

jonathan-mayer (changes requested):

jonathan-mayer (approved):
LGTMAWDTTOAYTD

Issue #221, opened 2025-05-20
Scale PodDisruptionBudgets Using Percentage Values


samuel-esp:
## Issue

PodDisruptionBudgets are currently unable to be scaled if the value of spec.maxUnavailable or spec.minAvailable is a percentage. We should implement a logic to handle percentage values in a way it could be used without modifying the main algorithm logic

It would be a plus if the proposed solution can be replicated also on PyKubeDownscaler without breaking the existing logic. Otherwise the two tools will have two different implementations/behavior but this won't be a problem

### Proposal

- Once we agree on a solution I'll edit this section and put it here

### Who Can Address the Issue

- Go dev

### Other Links/References

Related Issue On PyKubeDownscaler: caas-team/py-kube-downscaler#159

Issue #156, opened 2025-03-19
Restructure Website


jonathan-mayer:
### Problem to Solve

Currently the docs are a bit hard to follow. This is can at least partly be improved by restructuring the documentation.

### Further Details

This was originally proposed by @samuel-esp.

### Proposal

- [x] Introduction page
  - [ ] schema/diagram of the architecture
  - [x] briefly explain that scaling configuration can be given through cli, env, annotations on workload, namespace
- [x] explain the main concepts
  - [x] supported workload resources
  - [x] scopes
- [x] from the current scopes page, link to a page where we go in detail about the scope
  - [x] explain e.g. how to set up the scope
  - [ ] instead of linking to the values/annotations page, use [partial markdown](https://docusaurus.io/docs/markdown-features/react#importing-markdown) to have them on the scope page directly
- [x] extract all repo maintenance and developing docs/guides (Workflows documentation, Developing guides)
      to a new "contributing" sidebar

![Image](https://github.com/user-attachments/assets/e4e23cab-1003-43ff-90fc-7ec8c455025a)

### Other Links/References

https://docusaurus.io/docs/markdown-features/react#importing-markdown

PR #173, opened 2025-03-26
Refactor/website

samuel-esp:
## Motivation

As mentioned in issue #156, this pull request aims to improve the way the documentation on the site is presented so that it can be better understood by users

## Changes

Copying the to-do list made by @jonathan-mayer 

- Introduction page  
  - Schema/diagram of the architecture  
  - Brief explanation that scaling configuration can be provided through CLI, environment variables, annotations on workload, and namespace  

- Explanation of main concepts  
  - Supported workload resources  
  - Scopes  

- Detailed scope explanation linked from the current scopes page  
  - Instructions on setting up the scope  
  - Instead of linking to the values/annotations page, use [partial markdown](https://docusaurus.io/docs/markdown-features/react#importing-markdown) to include them directly on the scope page  

- Extraction of all repository maintenance and development documentation (Workflows documentation, Developing guides) into a new "Contributing" sidebar  

## Tests Done

- Unit Tests

## TODO

- [x] Introduction page
  - [x] schema/diagram of the architecture
  - [x] briefly explain that scaling configuration can be given through cli, env, annotations on workload, namespace
- [x] explain the main concepts
  - [x] supported workload resources
  - [x] scopes
- [x] from the current scopes page, link to a page where we go in detail about the scope
  - [x] explain e.g. how to set up the scope
  - [x] instead of linking to the values/annotations page, use [partial markdown](https://docusaurus.io/docs/markdown-features/react#importing-markdown) to have them on the scope page directly
- [x] extract all repo maintenance and developing docs/guides (Workflows documentation, Developing guides)
      to a new "contributing" sidebar

PR #222, opened 2025-05-24
Feat/scale podDisruptionBudgets with percentage values

samuel-esp:
## Motivation

As stated in #221 actually the GoKubeDownscaler can't scale down PodDisruptionBudgets with percentage values. This PR introduces a strategy to support them without compromising the existing logic

The following feature request was also made on PyKubeDownscaler (caas-team/py-kube-downscaler#160)

## Changes

- Added an interface to dynamically check which workload supports percentage values
- Enhanced logic for `getOriginalReplicas`, `setOriginalReplicas` to safely fetch percentage values
- Enhanced logic for `getMaxUnavailable`, `setMaxUnavailable`, `getMaxUnavailable`, `setMaxUnavailable`, `scaleDown`, `scaleUp` to safely get and update percentage values

## Tests Done

- Refactored Unit Tests

## TODO

- [x] Update Docs on #156

### DIFF
--- a/internal/pkg/scalable/errors.go
+++ b/internal/pkg/scalable/errors.go
@@ -40,15 +40,3 @@ func newOriginalReplicasUnsetError(reason string) error {
 func (o *OriginalReplicasUnsetError) Error() string {
 	return o.reason
 }
-
-type InvalidReplicaTypeError struct {
-	reason string
-}
-
-func newInvalidReplicaTypeError(reason string) error {
-	return &InvalidReplicaTypeError{reason: reason}
-}
-
-func (p *InvalidReplicaTypeError) Error() string {
-	return p.reason
-}
--- a/internal/pkg/scalable/replicaScaledWorkloads.go
+++ b/internal/pkg/scalable/replicaScaledWorkloads.go
@@ -59,7 +59,7 @@ func (r *replicaScaledWorkload) ScaleUp() error {
 func (r *replicaScaledWorkload) ScaleDown(downscaleReplicas values.Replicas) error {
 	downscaleReplicasInt32, err := downscaleReplicas.AsInt32()
 	if err != nil {
-		return newInvalidReplicaTypeError("failed to convert downscale replicas to int32")
+		return fmt.Errorf("failed to convert replicas to int32: %w", err)
 	}
 
 	originalReplicas, err := r.getReplicas()
--- a/internal/pkg/scalable/replicaScaledWorkloads_test.go
+++ b/internal/pkg/scalable/replicaScaledWorkloads_test.go
@@ -19,6 +19,7 @@ func TestReplicaScaledWorkload_ScaleUp(t *testing.T) {
 		originalReplicas     values.Replicas
 		wantOriginalReplicas values.Replicas
 		wantReplicas         values.Replicas
+		wantErr              error
 	}{
 		{
 			name:                 "scale up",
@@ -35,7 +36,7 @@ func TestReplicaScaledWorkload_ScaleUp(t *testing.T) {
 			wantReplicas:         values.AbsoluteReplicas(5),
 		},
 		{
-			name:                 "orignal replicas not set",
+			name:                 "original replicas not set",
 			replicas:             values.AbsoluteReplicas(0),
 			originalReplicas:     nil,
 			wantOriginalReplicas: nil,
@@ -47,6 +48,7 @@ func TestReplicaScaledWorkload_ScaleUp(t *testing.T) {
 			originalReplicas:     values.PercentageReplicas(50),
 			wantOriginalReplicas: nil,
 			wantReplicas:         values.AbsoluteReplicas(0),
+			wantErr:              &values.InvalidReplicaTypeError{},
 		},
 	}
 
@@ -56,7 +58,6 @@ func TestReplicaScaledWorkload_ScaleUp(t *testing.T) {
 
 			deployment := &replicaScaledWorkload{&deployment{&appsv1.Deployment{}}}
 			replicasInt32, _ := test.replicas.AsInt32()
-
 			_ = deployment.setReplicas(replicasInt32)
 
 			if test.originalReplicas != nil {
@@ -64,27 +65,24 @@ func TestReplicaScaledWorkload_ScaleUp(t *testing.T) {
 			}
 
 			err := deployment.ScaleUp()
-			if _, ok := test.originalReplicas.(values.PercentageReplicas); ok {
-				require.Error(t, err)
-				assert.Contains(t, err.Error(), "percentage")
+			var invalidReplicaTypeError *values.InvalidReplicaTypeError
 
+			if errors.As(test.wantErr, &invalidReplicaTypeError) {
+				assert.ErrorAs(t, err, &invalidReplicaTypeError)
 				return
 			}
 
 			require.NoError(t, err)
 
 			replicas, err := deployment.getReplicas()
-
-			if assert.NoError(t, err) {
-				assert.Equal(t, test.wantReplicas, replicas)
-			}
+			require.NoError(t, err)
+			assert.Equal(t, test.wantReplicas, replicas)
 
 			oringalReplicas, err := getOriginalReplicas(deployment)
+			var unsetErr *OriginalReplicasUnsetError
 
-			var originalReplicasUnsetErr *OriginalReplicasUnsetError
-
-			if ok := errors.As(err, &originalReplicasUnsetErr); !ok { // ignore getOriginalReplicas being unset
-				require.NoError(t, err) // Scaling set OrignialReplicas to faulty value
+			if !errors.As(err, &unsetErr) {
+				require.NoError(t, err)
 			}
 
 			assert.Equal(t, test.wantOriginalReplicas, oringalReplicas)
@@ -99,13 +97,16 @@ func TestReplicaScaledWorkload_ScaleDown(t *testing.T) {
 		name                 string
 		replicas             values.Replicas
 		originalReplicas     values.Replicas
+		downtimeReplicas     values.Replicas
 		wantOriginalReplicas values.Replicas
 		wantReplicas         values.Replicas
+		wantErr              error
 	}{
 		{
 			name:                 "scale down",
 			replicas:             values.AbsoluteReplicas(5),
 			originalReplicas:     nil,
+			downtimeReplicas:     values.AbsoluteReplicas(0),
 			wantOriginalReplicas: values.AbsoluteReplicas(5),
 			wantReplicas:         values.AbsoluteReplicas(0),
 		},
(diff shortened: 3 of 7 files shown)


internal/pkg/scalable/util.go
```diff
@@ -199,34 +199,50 @@ func isManagedByOwnerReference(workload Workload) bool {
...
+			return nil, &emptyStr, newPercentageValueNotAllowedError("error: percentage value not allowed for this workload type")
+		}
+
+		return nil, &originalReplicasString, nil
 	}
 
 	originalReplicas, err := strconv.ParseInt(originalReplicasString, 10, 32)
 	if err != nil {
-		return nil, fmt.Errorf("failed to parse original replicas annotation on workload: %w", err)
+		return nil, &emptyStr, fmt.Errorf("failed to parse original replicas annotation on workload: %w", err)
```
jonathan-mayer, line 239: we should just return nil instead of an pointer to an empty string

jonathan-mayer (changes requested):
In general I dont really like passing around both the int and the a whole string. I think a interface based typing would be way better. Additionally I would really want that you can set the downtime replicas to a percentage value, this would help bring back the consistency we strive(d) for with the GoKubeDownscaler and this implementation currently still lacks.

samuel-esp:
@jonathan-mayer I put some nolint just to ship this version quickly, I'll adjust with feedback. I implemented with 2 interfaces but i think I could have implemented also using something like 

```go
type ReplicaCount struct {
	isPercentage   Bool
	replicas int32
}
```

With interfaces I get many ireturn from the linters, maybe it's not the best approach in Go. It feels a bit java style

*There isn't still the logic to limit downscaler/downtime-replicas just to resources

jonathan-mayer (changes requested):
I didnt leave anything on all the errors.New and fmt.Errorf that were used to not deal with custom errors right now but we **need** to replace those.

Also remove the nolint nolintlint and just remove the linter that isnt required to be nolinted anymore from the nolint directive.

Sorry for all the review comments 🙃 If you want I can help refactor this.

internal/pkg/scalable/replicaCount.go
```diff
@@ -0,0 +1,50 @@
...
+	return errors.New("invalid value for ReplicaCount: must be int or percentage (e.g. 3 or 50%%)")
+}
+
+func (r *ReplicaCountValue) String() string {
+	if r.ReplicaCount == nil || *r.ReplicaCount == nil {
+		return ""
+	}
+
+	return (*r.ReplicaCount).String()
+}
```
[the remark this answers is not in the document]
samuel-esp: won't be complex to switch to an alias type here? since it wraps an interface type

samuel-esp:
I left some points opened and I still need to create all the custom error. Feel free to check out inside my code if you want to change something and have time, otherwise reviews are fine

internal/pkg/scalable/replicaCount.go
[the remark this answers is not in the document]
jonathan-mayer: although we should still replace the empty string with util.Undefined. (going to leave this review comment open for this)

internal/pkg/values/scope.go
```diff
@@ -45,24 +46,24 @@ func (s ScopeID) String() string {
 // NewScope gets a new scope with all values in an unset state.
 func NewScope() *Scope {
 	return &Scope{
-		DownscaleReplicas: util.Undefined,
+		DownscaleReplicas: &scalable.AbsoluteReplicas{Value: util.Undefined},
```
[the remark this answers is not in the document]
jonathan-mayer: left some review comments on the checks that need to be adjusted. 
couldnt leave one there but once we do refactor https://github.com/caas-team/GoKubeDownscaler/pull/222#discussion_r2108995814 we should also do scaledobjects.go:39

internal/pkg/values/scopeParser.go
```diff
@@ -201,23 +198,14 @@ func (s *Scope) GetScopeFromAnnotations( //nolint: nolintlint,funlen,gocognit,cy
...
-			downscaleReplicas = scalable.PercentageReplicas{Value: downscaleReplicasString}
-		} else {
-			//nolint: err113 // temp
-			err = fmt.Errorf("failed to parse %q annotation: invalid format %q", annotationDownscaleReplicas, downscaleReplicasString)
-			logEvent.ErrorInvalidAnnotation(annotationDownscaleReplicas, err.Error(), ctx)
+		var replicasVal ReplicasValue
 
+		if err = replicasVal.Set(downscaleReplicasString); err != nil {
+			logEvent.ErrorInvalidAnnotation(annotationDownscaleReplicas, err.Error(), ctx)
 			return err
```
jonathan-mayer:
```suggestion
                        err = fmt.Errorf("failed to parse %q annotation: %w", annotationDownscaleReplicas, err)
			logEvent.ErrorInvalidAnnotation(annotationDownscaleReplicas, err.Error(), ctx)
			return err
```

internal/pkg/values/scopeParser.go
```diff
@@ -199,18 +198,14 @@ func (s *Scope) GetScopeFromAnnotations( //nolint: funlen,gocognit,cyclop,gocycl
...
-			err = fmt.Errorf("failed to parse %q annotation: %w", annotationDownscaleReplicas, err)
+		if err = replicasVal.Set(downscaleReplicasString); err != nil {
 			logEvent.ErrorInvalidAnnotation(annotationDownscaleReplicas, err.Error(), ctx)
-
 			return err
 		}
 
-		// #nosec G115 // downscaleReplicas gets parsed as a 32 bit integer, so any errors that could be thrown here are already handled above
-		s.DownscaleReplicas = int32(downscaleReplicas)
+		s.DownscaleReplicas = *replicasVal.Replicas
```
jonathan-mayer, line 208: This does have to be adjusted, since the .Set method currently expects ReplicasValue.Replicas to be non-nil also we do have the var right there so it should be simple:
```go
		var replicasVal ReplicasValue

		if err = ReplicasValue{Replicas: &s.DownscaleReplicas}.Set(downscaleReplicasString); err != nil {
			logEvent.ErrorInvalidAnnotation(annotationDownscaleReplicas, err.Error(), ctx)
			return err
		}
```

internal/pkg/scalable/errors.go
```diff
@@ -52,3 +52,12 @@ func newInvalidReplicaTypeError(reason string) error {
...
 }
+
+type InvalidValueError struct {
+	reason string
+	value  string
+}
+
+func (i *InvalidValueError) Error() string {
+	return fmt.Sprintf("error: %q, got %s.", i.reason, i.value)
+}
```
jonathan-mayer, line 63: doesnt get used

internal/pkg/values/scope.go
```diff
@@ -89,29 +87,13 @@ func (s *Scope) CheckForIncompatibleFields() error { //nolint: nolintlint, cyclo
...
-				"downscale replicas must be a percentage between 0% and 100%",
-				pct.Value,
-			)
-		}
-	} else {
-		return newInvalidValueError(
-			"invalid downscale replicas type",
-			fmt.Sprintf("%T", s.DownscaleReplicas),
-		)
 	}
```
jonathan-mayer: we shouldnt need this at all anymore since we handle this in the parsing process. This actually also comes with the benifit of having the CheckForIncompatibleFields function actually only do that instead of their validity

internal/pkg/scalable/replicaScaledWorkloads_test.go
```diff
@@ -132,3 +133,31 @@ func TestReplicaScaledWorkload_ScaleDown(t *testing.T) {
...
+		t.Parallel()
+
+		deployment := &replicaScaledWorkload{&deployment{&appsv1.Deployment{}}}
+		_ = deployment.setReplicas(5)
+
+		err := deployment.ScaleDown(values.PercentageReplicas(50))
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "percentage")
+	})
+}
```
jonathan-mayer, line 163: I think this would be better as just another test in the respective ScaleUp and ScaleDown test functions

internal/pkg/scalable/replicaScaledWorkloads.go
```diff
@@ -54,28 +51,29 @@ func (r *replicaScaledWorkload) ScaleUp() error {
...
 // ScaleDown scales down the underlying replicaScaledResource.
 // nolint: err113 // dynamic errors
-func (r *replicaScaledWorkload) ScaleDown(downscaleReplicas ReplicaCount) error {
-	downscaleReplicasInt, ok := downscaleReplicas.(*AbsoluteReplicas)
-	if !ok {
-		return errors.New("invalid replica type: expected AbsoluteReplicas")
+func (r *replicaScaledWorkload) ScaleDown(downscaleReplicas values.Replicas) error {
+	if _, ok := downscaleReplicas.(values.AbsoluteReplicas); !ok {
+		return newInvalidReplicaTypeError("error: percentage value not allowed for this workload type")
 	}
```
jonathan-mayer: also here, we should let the error from AsInt32 handle this

internal/pkg/scalable/replicaScaledWorkloads.go
```diff
@@ -47,23 +53,29 @@ func (r *replicaScaledWorkload) ScaleUp() error {
...
 
-	err = r.setReplicas(downscaleReplicas)
+	err = r.setReplicas(downscaleReplicasInt.Value)
 	if err != nil {
 		return fmt.Errorf("failed to set replicas for workload: %w", err)
 	}
 
-	setOriginalReplicas(originalReplicas, "", r)
+	setOriginalReplicas(&AbsoluteReplicas{Value: originalReplicas}, r)
```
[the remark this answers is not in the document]
jonathan-mayer: yea, we can leave it like this for now. i also think this is likely to get refactored out with #157

jonathan-mayer (changes requested):

internal/pkg/values/scope.go
samuel-esp: refactored this test inside a new test class for `replicas.go`

internal/pkg/values/replicas_test.go
```diff
@@ -0,0 +1,162 @@
...
+				return
+			}
+
+			require.NotNil(t, replica)
+			assert.Equal(t, test.expected.String(), replica.String())
+		})
+	}
+}
+
+func TestReplicasValue_String(t *testing.T) {
```
jonathan-mayer, line 122: the stringer is only really used for debug logs so we shouldnt test for those

jonathan-mayer (changes requested):

jonathan-mayer (approved):
LGTM 👍

samuel-esp:
@jonathan-mayer in the afternoon I'll update the docs on #173, after that we can merge this one

samuel-esp:
> @jonathan-mayer in the afternoon I'll update the docs on #173, after that we can merge this one

Done!

PR #251, opened 2025-06-30
feat: add deadcode check

jonathan-mayer:
## Motivation

This pr adds a new pre commit hook that uses https://go.dev/blog/deadcode to check for unused/unreachable code.

We already have a linter that should handle this (golangci-lints unused check as well as staticcheck) but seemingly all unused linters do not take the `internal` folder into account. Which for us results in most functions not being checked since they are exported.

I had to make a simple pre commit hook for this myself since the only exisiting one i found was just not functional, not working in go projects with any folder structure and not even handling the command output at all.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,7 +24,7 @@ repos:
     hooks:
       - id: go-deadcode
   - repo: https://github.com/rbubley/mirrors-prettier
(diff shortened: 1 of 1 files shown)


JTaeuber (approved):
Sure ¯\\\_(ツ)\_/¯

PR #275, opened 2025-08-19
Refactor/installation docs

samuel-esp:
## Motivation

Right now the installation guide it's not very explicit on Cluster-Wide mode vs Constrained mode. Users can find out about these options reading carefully the `values.yaml` as pointed out in the last subsection of the page but sometimes this could be overlooked if the information is not immediately made available to the user ([py-kube-downscaler installation docs](https://github.com/caas-team/py-kube-downscaler?tab=readme-ov-file#installation)). Since these modes are important to know before starting the installation process, I think it would better to explicitly create a dedicated subsection explaining them (and how to set them in the installation process) just before the installation sub section

## Changes

- added a subsection inside the installation guide page of the documentation that allows users to better understand how to activate the constrainedNamespace option

## Tests Done

- Not needed

### DIFF
--- a/website/content/guides/0 - getting-started/0 - Installation.mdx
+++ b/website/content/guides/0 - getting-started/0 - Installation.mdx
@@ -19,6 +19,24 @@ In order to get started installing the GoKubeDownscaler using our Helm Chart you
 
 Also make sure your context is set to the right cluster and namespace where you want to deploy the GoKubeDownscaler.
 
+## Constrained Installation vs Cluster-Wide Installation
+
+As described in the [Architecture](ref:docs-architecture) section, the GoKubeDownscaler can be installed in two different ways:
+
+- **Constrained Installation**: The GoKubeDownscaler is only allowed to operate on a set of namespaces.
+  This is useful if you don't have the
+  necessary rights to make a cluster-wide installation.
+- **Cluster-Wide Installation**: The GoKubeDownscaler is allowed to operate
+
+In order to install the GoKubeDownscaler in a constrained way you must set the `constrainedNamespaces` value.
+This can be done in two ways:
+
+- by passing the `--set constrainedNamespaces=[namespace1,namespace2,...]` argument to the `helm install` or `helm template` command shown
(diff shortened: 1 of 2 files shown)


website/content/guides/0 - getting-started/0 - Installation.mdx
```diff
@@ -19,6 +19,24 @@ In order to get started installing the GoKubeDownscaler using our Helm Chart you
 
 Also make sure your context is set to the right cluster and namespace where you want to deploy the GoKubeDownscaler.
 
+## Constrained Installation vs Cluster-Wide Installation
+
+As described in the [Architecture](ref:docs-architecture) section, the GoKubeDownscaler can be installed in two different ways:
```
JTaeuber, line 24: I think you probably mean this with the architecture section.
```suggestion
As described in the [introduction page](ref:docs-introduction#installation-and-architecture), the GoKubeDownscaler can be installed in two different ways:
```

website/content/guides/0 - getting-started/0 - Installation.mdx
```diff
@@ -19,6 +19,24 @@ In order to get started installing the GoKubeDownscaler using our Helm Chart you
...
+
+- **Constrained Installation**: The GoKubeDownscaler is only allowed to operate on a set of namespaces.
+  This is useful if you don't have the
+  necessary rights to make a cluster-wide installation.
+- **Cluster-Wide Installation**: The GoKubeDownscaler is allowed to operate
```
JTaeuber, line 29: I think something might be missing here.
Maybe something along the lines of: `The GoKubeDownscaler is allowed to operate on all the namespaces of the entire cluster`

JTaeuber (changes requested):

website/content/guides/0 - getting-started/0 - Installation.mdx
samuel-esp: Thanks for pointing that out, didn't correctly tab out the ide comment suggestion

PR #413, opened 2026-03-08
Feat/Add Service, Ingress, Gateway Resources

### LINKED ISSUE #299: Implement LoadBalancers Scaling: Services And Ingresses
## Issue

In an Internal Development Platform (IDP) LoadBalancers could be a major cost factor (the average LoadBalancer monthly cost across cloud providers is around 18$). Usually LoadBalancers are provisioned using one of these strategies:

- **LoadBalancers/Ingresses Per Namespace**: some organizations may choose to give users enough freedom to provision their own set of LoadBalancers or Ingresses 
- **Centralized LoadBalancers/Ingresses**: some organizations may choose to create a fixed set of LoadBalancers/Ingresses to redirect traffic towards different namespaces

Scaling LoadBalancers could be beneficial for both scenarios

### Proposal

Implement `services` and `ingresses`  and as a scalable resource. 

- `services` should scale down from LoadBalancer to ClusterIP
- `ingresses` should scale down from "real" IngressClass to "fake" IngressClass

It could be needed to address special cases, in some cases some types of LoadBalancer should be excluded while others not

> 
>  For example in AWS NLB and ELB are often treated differently:
> 
> - ELB doesn't have a static IP -> user shouldn't reley on IPs
> - NLB does have a static IP -> user may rely on IPs
> 
> Architectural components that directly reference NLB static IPs could not tolerate such a downscaling 

Users should be informed that CNAMEs could be automatically reconfigured using External DNS Kubernetes plugin

> 
> Before Downscaling: example.com -> load-balancer.abc.cloudprovider.com
> After Downscaling: example.com -> load-balancer.def.cloudprovider.com (without manual intervention)

### Who Can Address the Issue

- Go Developers

### Other Links/References

[External DNS](https://github.com/kubernetes-sigs/external-dns)
[EKS Service Controller](https://cloud-provider-aws.sigs.k8s.io/service_controller/)
[EKS ALB Controller](https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/)
[AKS LoadBalancers](https://learn.microsoft.com/en-us/azure/aks/internal-lb?tabs=set-service-annotations)
[GKE LoadBalancers](https://cloud.google.com/kubernetes-engine/docs/concepts/service-load-balancer)

samuel-esp:
## Motivation

As part of the feature request highlighted in #299, Services are an integral part of the costs incurred when using a cloud Kubernetes distribution like EKS, AKS, or GKE. This PR adds the ability to scale Services of type LoadBalancer.

How it works for Services:
- During downtime: `Services` of type `LoadBalancers` are converted to `ClusterIP`
- During uptime: `Services` of type `ClusterIP` which were previously of type `LoadBalancer` are converted again to `LoadBalancer`

This PR supports also the possibility to distinguish between AWS LoadBalancer services. Users can choose if they want to scale ELB Services or NLB Services independently.

How it works for Ingresses:
- During downtime: `Ingresses` are scaled to a downscaler reserved `IngressClassName`. This way, the original Ingress controller will no longer manage this resource and will remove it from the CloudProvider cloud.
- During uptime: `Ingresses` which were previously downscaled are restored with their original `IngressClassName`. The original Ingress controller will start managing it again

How it works for Gateways:
- During downtime: `Gateways` are scaled to a downscaler reserved `GatewayClassName`. This way, the original Gateway controller will no longer manage this resource and will remove it from the CloudProvider cloud.
- During uptime: `Gateways` which were previously downscaled are restored with their original `IngressClassName`. The original Ingress controller will start managing it again

## Changes

- Implemented `service`, `ingress` and `gateway` as new resources
- Implemented a new type of replicas: `StatusReplicas`. Refactored tests for replicas
- Implemented Helm Logic
- Add permissions in the chart
- Add documentation related to this PR 

## Tests Done

- Unit Tests
- Live Tests

## TODO

- [x] If possible, refactor these resources to a new interface called "classScaledWorkloads" or "fieldScaledWorkloads" since they share behavior logic
- [x] Further Tests


internal/pkg/scalable/workload.go
```diff
@@ -71,6 +77,11 @@ func ParseWorkloadFromRawObject(resource string, rawObject []byte) (Workload, er
 		"stack":                   parseStackFromBytes,
 		"prometheus":              parsePrometheusFromBytes,
 		"autoscalingrunnerset":    parseAutoscalingRunnerSetFromBytes,
+		"services":                parseServiceFromBytes,
+		"awsnlbservice":           parseServiceFromBytes,
+		"awselbservice":           parseServiceFromBytes,
+		"ingresses":               parseIngressesFromBytes,
+		"gateways":                parseGatewaysFromBytes,
```
jonathan-mayer, line 84: I think these need to be singular

deployments/chart/templates/downscaleringressclass.yaml
```diff
@@ -0,0 +1,9 @@
+{{- /* Only deploy when the user explicitly enables deployment AND the resource is included in includedResources */ -}}
+{{- if and .Values.deployExtraResources.deployIngressClass (include "go-kube-downscaler.deployIngressClass" .) -}}
```
jonathan-mayer, line 2:
```suggestion
{{- if and .Values.deployExtraResources.ingressClass (include "go-kube-downscaler.deployIngressClass" .) -}}
```

internal/pkg/values/replicas.go
```diff
@@ -39,6 +40,18 @@ func (p PercentageReplicas) AsIntStr() intstr.IntOrString {
 	return intstr.FromString(strconv.Itoa(int(p)) + "%")
 }
 
+type StatusReplicas string
```
jonathan-mayer, line 43: i feel like StringReplicas could be a better name

internal/pkg/scalable/valueScaledWorkloads.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// ScaleDown scales down the underlying valueScaledResource.
+func (v *valueScaledWorkload) ScaleDown(_ values.Replicas) (*metrics.SavedResources, error) {
+	currentState, targetScaleDownState, err := v.getValue()
+	if err != nil {
+		return metrics.NewSavedResources(0, 0), err
+	}
+
+	if currentState == targetScaleDownState {
```
jonathan-mayer, line 66: this currently checks if both values.Replicas are pointing to the same address. we should get the string here and check it

internal/pkg/scalable/valueScaledWorkloads.go
```diff
@@ -0,0 +1,95 @@
...
+	"errors"
+	"fmt"
+	"log/slog"
+
+	"github.com/caas-team/gokubedownscaler/internal/pkg/metrics"
+	"github.com/caas-team/gokubedownscaler/internal/pkg/values"
+	"github.com/wI2L/jsondiff"
+)
+
+// valueScaledResource provides all the functions needed to scale a resource which is scaled by setting a suspend field.
```
jonathan-mayer, line 14: copy paste error

website/content/docs/3 - helm-chart/0 - Components.mdx
```diff
@@ -101,3 +101,19 @@ The [validatingadmissionpolicyremoval.yaml](repo:deployments/chart/templates/val
 ValidatingAdmissionPolicy that blocks the removal of any `downscaler/` annotations from namespaces or workloads
 if unauthorized users or service accounts tried removing them.
 When deployed, the policy is bound to every namespace.
+
+## IngressClass
+
+The [downscalergatewayclass.yaml](repo:deployments/chart/templates/downscaleringressclass.yaml) file creates a GatewayClass that
```
jonathan-mayer, line 107:
```suggestion
The [downscaleringressclass.yaml](repo:deployments/chart/templates/downscaleringressclass.yaml) file creates a IngressClass that
```

website/content/docs/2 - downscaler/3 - Workload Types.mdx
```diff
@@ -106,3 +106,46 @@ Scales by setting the replica count to the [downscale replicas](ref:docs-values#
...
+set to `nlb` are scaled.
+
+### Ingresses
+
+- id: ingresses
+- resource: ingresses.v1.networking.k8s.io
+
+Scales by setting the IngressClassName of the Gateway object to `downscaler-ingress-class`.
+In a constrained installation, the GatewayClass object with the name `downscaler-gateway-class` must
+be provisioned by a cluster administrator having the necessary cluster-wide permissions
```
jonathan-mayer, line 142:
```suggestion
Scales by setting the IngressClassName of the Ingress object to `downscaler-ingress-class`.
In a constrained installation, the IngressClass object with the name `downscaler-ingress-class` must
be provisioned by a cluster administrator having the necessary cluster-wide permissions
```

jonathan-mayer (changes requested):

internal/pkg/scalable/valueScaledWorkloads.go
samuel-esp: Hey @jonathan-mayer I'll leave this open for a double check on your side,  `currentState == targetScaleDownState` already performs a value comparison. If I understood correctly what you meant

jonathan-mayer: I dont think it actually does, its trying to compare two values.Replicas interfaces, wouldnt that compare their pointers

samuel-esp: Isn't the same situation specified [here](https://go.dev/ref/spec#Comparison_operators:~:text=Interface%20types%20that%20are%20not%20type%20parameters%20are%20comparable%2E%20Two%20interface%20values%20are%20equal%20if%20they%20have%20identical%20dynamic%20types%20and%20equal%20dynamic%20values%20or%20if%20both%20have%20value%20nil%2E)? 

Look also this [playground](https://go.dev/play/p/Qzq4u9o65j8). This might be an edge case all `values.Replicas` are primitive types

jonathan-mayer (dismissed):

jonathan-mayer (approved):

PR #433, opened 2026-03-30
fix(helm): correct MutatingAdmissionPolicy matchConditions CEL expressions

gaetanars:
## Summary

All `matchConditions` CEL expressions in `mutatingadmissionpolicy.yaml` used `oldObject != null` where `oldObject == null` was intended.

The condition is part of an OR clause that determines whether a `downscaler/` annotation is being **added for the first time**:

```
oldObject == null                                              → CREATE (no prior object)
|| !has(oldObject.metadata.annotations)                       → prior object had no annotations
|| !(l in oldObject.metadata.annotations)                     → annotation is new
|| object.metadata.annotations[l] != oldObject.metadata.annotations[l]  → value changed
```

With `!= null`, the first clause evaluated to `true` only when `oldObject` was present (UPDATE operations), making the OR effectively always true on UPDATE — but this branch was supposed to catch CREATE operations (new objects with no prior state). Correcting to `== null` restores the intended semantics.

## Impact

Without this fix, the `MutatingAdmissionPolicy` (Kubernetes ≥ 1.34) would not correctly block unauthorized `downscaler/` annotation additions on freshly created resources (CREATE requests). UPDATE protection was unaffected.

## Post-Deploy Monitoring & Validation

No additional operational monitoring required: template-only fix with no runtime behavior changes outside `MutatingAdmissionPolicy` enforcement (opt-in feature, requires K8s ≥ 1.34 and `annotationsCompliance.mutateUnauthorizedAnnotationsAddition: true`).

---

🤖 Generated with Claude Sonnet 4.6 (200K context) via [Claude Code](https://claude.com/claude-code) + Compound Engineering v2.40.0

### DIFF
--- a/deployments/chart/templates/mutatingadmissionpolicy.yaml
+++ b/deployments/chart/templates/mutatingadmissionpolicy.yaml
@@ -25,29 +25,29 @@ spec:
     {{- end }}
   matchConditions:
     - name: kube-downscaler-webhook-sa
-      expression: 'has(object.metadata.annotations) && object.metadata.annotations.exists(l, l.startsWith("downscaler/") && (oldObject != null || !has(oldObject.metadata.annotations) || !(l in oldObject.metadata.annotations) || object.metadata.annotations[l] != oldObject.metadata.annotations[l])) && request.userInfo.username != "system:serviceaccount:{{ .Release.Namespace }}:{{ include "go-kube-downscaler.webhookController.fullname" . }}"'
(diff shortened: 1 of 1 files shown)


samuel-esp (approved):
LGTM. Thanks for catching this! Writing these expression was a bit difficult as you may imagine and I missed this error 😅 . Tested in my cluster and I managed to replicate the issue and evaluate the PR, seems all good

PR #434, opened 2026-03-30
feat(scalable): add Strimzi KafkaConnect, KafkaMirrorMaker2, KafkaBridge support

gaetanars:
## Summary

- Adds downscaling support for three [Strimzi Kafka operator](https://strimzi.io/) workload types: `KafkaConnect`, `KafkaMirrorMaker2`, and `KafkaBridge`
- Uses the existing controller-runtime unstructured client — no new Go module dependencies
- Gracefully handles absent Strimzi CRDs via `apimeta.IsNoMatchError` (scan loop continues normally)
- Enabled via `--include-resources kafkaconnects`, `kafkamirrormaker2s`, `kafkabridges` (opt-in, consistent with other CRD-backed workloads)

### New: `internal/pkg/scalable/strimzi.go`
Single file implementing the `strimziWorkload` wrapper for all three types. Scaling operates on `spec.replicas`. Handles `float64`/`int64` JSON type ambiguity from the API server (unstructured maps decode numbers as `float64`). Uses `DeepCopy()` to avoid shallow map copies.

### Updated: `internal/pkg/scalable/workload.go`
Registers the three resource IDs in `resourceFuncMap` (scanner) and `parseWorkloadFuncMap` (webhook).

### Updated: `deployments/chart/templates/_helpers.tpl`
- RBAC rules (`go-kube-downscaler.permissions`) for `kafka.strimzi.io` group
- `MutatingWebhookConfiguration` rules (`go-kube-downscaler.webhookresources`)
- `MutatingAdmissionPolicy` rules (`go-kube-downscaler.annotationsCompliance`)

### Updated: `deployments/chart/values.yaml`
Adds three commented-out entries to `includedResources`.

### Updated: `website/content/docs/2 - downscaler/3 - Workload Types.mdx`
Documents the three new workload types.

## Testing

- Unit tests in `strimzi_test.go` covering:
  - `getReplicas` with `float64` (API-server JSON), `int64`, and absent `spec.replicas`
  - `setReplicas` round-trip
  - `Copy()` deep-copy isolation (mutating original doesn't affect copy)
  - GVK correctness for all three types
- `go test ./...` passes

## Post-Deploy Monitoring & Validation

- **Logs to watch**: `"strimzi CRD not found in cluster, skipping"` → expected when Strimzi is not installed; `"scaling down kafkaconnect"` → expected when enabled and in downscale window
- **Failure signal**: `"failed to get KafkaConnects"` errors → check RBAC (cluster role may need updating after Helm upgrade)
- **Validation**: enable `kafkaconnects` in `includedResources`, verify replica count reaches downscale value during window and recovers on upscale
- **No impact** when Strimzi CRDs are absent — resource types remain opt-in

---

🤖 Generated with Claude Sonnet 4.6 (200K context) via [Claude Code](https://claude.com/claude-code) + Compound Engineering v2.40.0


samuel-esp:
Hey @gaetanars thank you for contributing to the GoKubeDownscaler! If you don't mind we'll proceed merging this PR over the one you made in the PR repository since from now on we are promoting this tool over the old one. Give us some day so we can full review this, we can release the version including Strimzi operator in `1.4.0`

gaetanars:
Hey @samuel-esp 👋 friendly nudge on this one — happy to address any review feedback whenever you have a moment. Let me know if anything blocks the path to `1.4.0`, I'll happily rebase or adjust as needed. Thanks again for the warm reception!

samuel-esp:
Hey @gaetanars sorry for the late response on this, I've been a bit busy on other tasks the last couple of weeks. I'll publish the review by the next couple of days! So you can address the changes. Of course still planned for `1.4.0`, every resource addition is more than accepted for us 😄 

We plan to release `1.3.4` quite soon with some behavior fixes and then `1.4.0`. `1.4.0` Will have strimzi resources (from this PR), load balancers scaling, open kruise scaling and a new timestamp 😄

internal/pkg/scalable/strimzi.go
```diff
@@ -0,0 +1,222 @@
...
+	kafkaConnectGVK = schema.GroupVersionKind{
+		Group: "kafka.strimzi.io", Version: "v1beta2", Kind: "KafkaConnect",
+	}
+	kafkaMirrorMaker2GVK = schema.GroupVersionKind{
+		Group: "kafka.strimzi.io", Version: "v1beta2", Kind: "KafkaMirrorMaker2",
+	}
+	kafkaBridgeGVK = schema.GroupVersionKind{
+		Group: "kafka.strimzi.io", Version: "v1beta2", Kind: "KafkaBridge",
+	}
+)
```
samuel-esp: Even though the resources are strongly correlated by ApiGroup. To keep the convention used in the repo it would be better to create 3 separate classes. 1 for each resource (`KafkaBridge`, `KafkaMirrorMaker`, `KafkaConnect`)

samuel-esp (changes requested):
Overall the PR is good, it just needs the per-resource class refactoring.

Unfortunately here it is not possible to use a Clientset or a simple Client since Strimzi is a Java project and doesn't expose nothing natively in Go. The use of `Unstructured` seems the only viable way to me in this case.

gaetanars:
Hey @samuel-esp, thanks for the review! Pushed two follow-up commits addressing the feedback:

- [`bf46fcd`](https://github.com/caas-team/GoKubeDownscaler/pull/434/commits/bf46fcd2d45bfaa8842462b160b27a03a39a5058) — **`refactor(scalable): split Strimzi workload into per-resource files`**: replaces the shared `strimziWorkload` with three concrete types (`kafkaConnect`, `kafkaMirrorMaker2`, `kafkaBridge`), each in its own file with focused tests, matching the per-resource convention used by `rollouts.go`, `scaledobjects.go`, `stacks.go`. Behaviour-preserving; `Unstructured` kept as agreed.
- [`f3d34fa`](https://github.com/caas-team/GoKubeDownscaler/pull/434/commits/f3d34fa424fd8b81153f5c9a1fac4522c562330d) — **`feat(scalable): migrate Strimzi resources to kafka.strimzi.io/v1`**: while reworking this I noticed Strimzi 0.49 promoted these CRDs to a stable `v1` API and 1.0.0 removed `v1beta2` entirely, so the GVKs/tests/docs have been moved to `v1`. This is a breaking change for clusters running Strimzi `< 0.49` — happy to drop or split this commit out of the PR if you'd prefer to land it separately, just let me know.

Tests still green locally (`go test ./internal/pkg/scalable/...`). Let me know if you'd like further adjustments before merging into `1.4.0` 🙏

samuel-esp:
@gaetanars code looks good to me, before approving the PR we'll wait some PRs to be merged before 1.4.0 (so we avoid accidental merge). I think it may need a rebase soon (but I think also it needs a rebase right now because I see tests failing for dependencies)

gaetanars:
Heads up: the `run-pre-commit` CI failure on this PR is unrelated to the Strimzi changes — it's a checksum mismatch in the `golangci-lint` install script (the `master` branch of `golangci/golangci-lint` is no longer maintained, see [this comment](https://github.com/golangci/golangci-lint/issues/6540#issuecomment-4360011730)).

I opened #450 with the recommended fix. Once that lands I'll rebase this branch on top to get a green CI here.

gaetanars:
Hey @samuel-esp — rebased on the latest `main` (includes 1.3.4) and the CI is now green. Ready for your final review whenever you are!

samuel-esp:
Hey @gaetanars, thank you for rebasing! Been busy weeks, I'll review them in the next few days!

internal/pkg/scalable/kafkaconnects.go
```diff
@@ -0,0 +1,169 @@
...
+
+	if !found {
+		return nil, newNoReplicasError(k.GetKind(), k.GetName())
+	}
+
+	switch v := val.(type) {
+	case int64:
+		return values.AbsoluteReplicas(int32(v)), nil //nolint:gosec // temporary in-place conversion
+	case float64:
+		return values.AbsoluteReplicas(int32(v)), nil
```
samuel-esp: I would avoid in-place conversion; you could define a function in `util.go` to handle these conversions and then use that function here before submitting to absolute replicas. If you find/have a better approach for this, I'm absolutely open

internal/pkg/scalable/kafkamirrormaker2s.go
```diff
@@ -0,0 +1,169 @@
...
+
+	if !found {
+		return nil, newNoReplicasError(k.GetKind(), k.GetName())
+	}
+
+	switch v := val.(type) {
+	case int64:
+		return values.AbsoluteReplicas(int32(v)), nil //nolint:gosec // temporary in-place conversion
+	case float64:
+		return values.AbsoluteReplicas(int32(v)), nil
```
samuel-esp: Same comment about handling conversion as above

samuel-esp (changes requested):
Hey @gaetanars, I'm very sorry about the delay in reviewing this, the past weeks were really complicated for me and I didn't find any time. I've just suggested a small correction about conversions because in my opinion it would be better to avoid in place conversions as the linter says. Otherwise it looks good to me. 

Other question, did you also have chance to live test this in a test cluster (a minikube one is fine)? Anyway if that is not the case don't worry, It's not a prerequisite because, just in case, I will try to test it by myself before making the next release

gaetanars:
Hey @samuel-esp, thanks for the review and no worries about the delay!

Both points are now addressed:

1. **Per-resource classes** — already done in a previous commit ([b4c7f25](https://github.com/caas-team/GoKubeDownscaler/commit/b4c7f25)): `strimzi.go` was split into `kafkaconnects.go`, `kafkamirrormaker2s.go`, and `kafkabridges.go`.

2. **Conversion helper** — just pushed: added `unstructuredReplicasToInt32` in `util.go` and updated the three `getReplicas()` functions to use it instead of duplicating the inline `switch float64/int64` ([eaf20cb](https://github.com/gaetanars/GoKubeDownscaler/commit/eaf20cb)).

I haven't had the chance to live-test on a real cluster yet — happy to do so if you'd like, or let me know if you want to test it on your side before the release.

samuel-esp:
Thanks a lot for the effort @gaetanars. Yes, if you have a chance to test it on a live cluster you would save us a bit of time 😄. Of course do it only if you have enough free time, we won't force anyone to do so.

Another thing to help us: leave the last commit always with green continuous integration checks, this helps us understanding if everything in the code is ok (formatting, linters and tests). If you see an automatic commit from the bot, you can always push an empty commit after that just to trigger continuous integration once again (or you could setup pre-commit in your local env and test with it before committing, this might be a bit complicated if you are working on a Windows machine)

internal/pkg/scalable/util.go
```diff
@@ -288,3 +290,17 @@ func derefInt32(p *int32, def int32) int32 {
...
+func unstructuredReplicasToInt32(val any) (int32, bool) {
+	switch v := val.(type) {
+	case int64:
+		return int32(v), true //nolint:gosec // replica counts fit in int32
+	case float64:
+		return int32(v), true
+	default:
+		return 0, false
+	}
+}
```
samuel-esp, line 321: This might be a bit safer for conversion, even though it's almost an edge case. Better to cover it in any case

```suggestion
func unstructuredReplicasToInt32(val any) (int32, bool) {
	switch v := val.(type) {
	case int64:
		if v < math.MinInt32 || v > math.MaxInt32 {
			return 0, false
		}
		return int32(v), true

	case float64:
		if v != math.Trunc(v) {
			return 0, false
		}
		if v < math.MinInt32 || v > math.MaxInt32 {
			return 0, false
		}
		return int32(v), true

	default:
		return 0, false
	}
}
```

samuel-esp (changes requested):
Just this small conversion safeguard

samuel-esp (approved):
LGMT

PR #439, opened 2026-04-08
Feat/Added DirectionalTimeSpan Logic (From/Until Timespan)

### LINKED ISSUE #320: Add until and from timespans
### Problem to Solve

Currently you cannot have a timespan based value be true after or until a specific time.

### Proposal

- [ ] add a new timespan type that holds a single time.Time and a bool or an enum defining if it will be evaluated as `<time` or `>=time`
- [ ] evaluate the best string representation for this timespan type. (maybe `until<RF3339-Timestamp>` and `from<RF3339-Timestamp>` (we could also have the words be `to` and `after`) or something else entirely)

- [ ] update the docs to reflect the changes:
  - [ ] add to changes in migration guide
  - [ ] add docs for arguments in timespan type configuration page

samuel-esp:
## Motivation

As pointed here #320, currently the downscaler doesn't have a dedicated Timespan which is always evaluated as `true` before or after a certain date. This PR adds the possibility to specify such a timespan. This timespan only supports RFC3339 specification as stated in the enhancement request

## Changes

- Added SingleTimeSpan (we could also find a better name)
- Added implementation in timespan.go
- Added Unit Tests  

## Tests Done

- Unit Tests

## TODO

- [x] Live Tests
- [x] Docs


samuel-esp:
Live tests and docs are ok

jonathan-mayer (dismissed):

samuel-esp:
@jonathan-mayer I've rebased this to address the latest changes in the repo

jonathan-mayer (approved):
idk why i never approved this, i had already reviewed the timespan.go, just not the docs

Issue #451, opened 2026-05-11
Helm chart: `constrainedNamespaces` silently no-ops unless users override `excludedNamespaces`, and overriding to `[]` excludes everything


gaetanars:
## Summary

The combination `constrainedNamespaces` + `excludedNamespaces` in the Helm chart has a sharp edge that produces a downscaler that runs healthy but scales nothing. Two issues compound:

1. **Design**: scoping the downscaler to a namespace via `--namespace` (CLI) / `constrainedNamespaces` (Helm) does not take precedence over `--exclude-namespaces`. When the included namespace is also in the default exclude list (e.g. `kube-system`), the downscaler scopes its API access to that namespace and then immediately filters all of its workloads back out.
2. **Bug**: trying to "clear" the exclude list from values with `excludedNamespaces: []` produces an env var `EXCLUDE_NAMESPACES=''` whose parsed value is a regex `""` that matches every namespace — so *all* workloads are filtered out.

The net effect: a user who legitimately wants the downscaler to manage `kube-system` only must override `excludedNamespaces`, and the most natural way to override it (`[]`) silently disables the downscaler completely.

## Repro

`values.yaml`:

```yaml
constrainedNamespaces:
  - kube-system
excludedNamespaces: []
```

Resulting cluster state:

- `Role`/`RoleBinding` created in `kube-system` only (chart switches from `ClusterRole` to `Role`, see `deployments/chart/templates/role.yaml` and the guard at `deployments/chart/templates/clusterrole.yaml:1`).
- Deployment receives `--namespace=kube-system` (`deployments/chart/templates/deployment.yaml:51-53`).
- ConfigMap renders `EXCLUDE_NAMESPACES: ''` (`deployments/chart/templates/configmap.yaml:7`) — the key is always emitted, even for an empty list.

At runtime:

- `os.LookupEnv(\"EXCLUDE_NAMESPACES\")` returns `ok=true` (env var exists, value is empty), so `RegexList.Set(\"\")` is called and the Go default `{kube-system, kube-downscaler}` (`internal/pkg/util/config.go:45`) is overwritten.
- `RegexList.Set(\"\")` does `strings.Split(\"\", \",\")` → `[\"\"]` → `regexp.Compile(\"\")` → a regex that matches every string (`internal/pkg/util/regexList.go:11-27`).
[...]

PR #452, opened 2026-05-11
fix(chart): omit EXCLUDE_NAMESPACES when excludedNamespaces is empty

gaetanars:
## Summary

Fixes part B of #451. Setting `excludedNamespaces: []` in `values.yaml` currently renders `EXCLUDE_NAMESPACES: ''` in the ConfigMap. Because the env var is *set* (with an empty value), the binary calls `RegexList.Set(\"\")`, which produces a regex `\"\"` that matches every string — silently excluding all namespaces and disabling the downscaler.

This PR guards the ConfigMap key with `{{- with .Values.excludedNamespaces }}` so an empty list falls back to the Go default (`kube-system`, `kube-downscaler`) instead of being rendered as an empty env var.

## Changes

- `deployments/chart/templates/configmap.yaml` — only emit `EXCLUDE_NAMESPACES` when the list is non-empty.
- `website/content/docs/3 - helm-chart/2 - values/16 - excludedNamespaces.mdx` — document the behavior and how to truly exclude nothing (`[\"$^\"]`).

## Verification

\`\`\`console
$ helm lint deployments/chart/
1 chart(s) linted, 0 chart(s) failed

$ helm template test deployments/chart/ | grep EXCLUDE_NAMESPACES
  EXCLUDE_NAMESPACES: 'kube-downscaler,kube-system'

$ helm template test deployments/chart/ --set-json 'excludedNamespaces=[]' | grep EXCLUDE_NAMESPACES
# (no output — key is omitted, Go defaults apply)
\`\`\`

## Scope

This is the chart-side fix (B) from #451. The companion design fix (A) — making `--namespace` take precedence over `--exclude-namespaces` so `constrainedNamespaces` works without touching `excludedNamespaces` — is intentionally left to a separate discussion in that issue.

Refs #451

### DIFF
--- a/deployments/chart/templates/configmap.yaml
+++ b/deployments/chart/templates/configmap.yaml
@@ -4,7 +4,9 @@ metadata:
   name: {{ .Values.configMap.name }}
   namespace: {{ .Release.Namespace }}
 data:
-  EXCLUDE_NAMESPACES: '{{- join "," .Values.excludedNamespaces }}'
+  {{- with .Values.excludedNamespaces }}
+  EXCLUDE_NAMESPACES: '{{ join "," . }}'
+  {{- end }}
   {{- if .Values.configMap.extraConfig }}
   {{- tpl .Values.configMap.extraConfig . | nindent 2 }}
   {{- end }}
--- a/deployments/chart/templates/configmap.yaml
+++ b/deployments/chart/templates/configmap.yaml
@@ -4,9 +4,7 @@ metadata:
   name: {{ .Values.configMap.name }}
   namespace: {{ .Release.Namespace }}
 data:
-  {{- with .Values.excludedNamespaces }}
-  EXCLUDE_NAMESPACES: '{{ join "," . }}'
-  {{- end }}
+  EXCLUDE_NAMESPACES: '{{ join "," (coalesce .Values.excludedNamespaces (list .Release.Namespace "kube-system")) }}'
   {{- if .Values.configMap.extraConfig }}
   {{- tpl .Values.configMap.extraConfig . | nindent 2 }}
   {{- end }}
--- a/website/content/docs/3 - helm-chart/2 - values/16 - excludedNamespaces.mdx
+++ b/website/content/docs/3 - helm-chart/2 - values/16 - excludedNamespaces.mdx
@@ -41,7 +41,8 @@ Will result in all workloads in the namespaces `kube-downscaler`, `kube-system` 
 
 :::note
 
-Setting `excludedNamespaces: []` does not clear the exclusion list. The chart omits the `EXCLUDE_NAMESPACES` env var when the list is empty, so the binary falls back to its built-in default (`kube-system`, `kube-downscaler`). To truly exclude nothing, set a regex that cannot match any namespace name, such as `["$^"]`.
+Setting `excludedNamespaces: []` does not clear the exclusion list.
+The chart omits the `EXCLUDE_NAMESPACES` env var when the list is empty, so the binary falls back to its built-in default (`kube-system`, `kube-downscaler`). To truly exclude nothing, set a regex that cannot match any namespace name, such as `["$^"]`.
 
 :::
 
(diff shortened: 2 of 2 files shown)


samuel-esp:
I think this could be a good change, but I'll leave the default exclusion only to the release namespace itself because we can't know if the user wants truly to exclude everything in kube-system. @JTaeuber @jonathan-mayer what do you think about that?

samuel-esp:
I'll correct myself: actually I forgot there is this default already in place inside the code which is the one you are referring to

https://github.com/caas-team/GoKubeDownscaler/blob/73643280bc837faab000739cd92257a6dd77b495/internal/pkg/util/config.go#L45

For now, I’d say keeping these default values is technically the right choice. While `kube-system` is universal `kube-downscaler` may seem unnecessary here, since users can install the chart into a namespace with a different name, however GoKubeDownscaler can also be run as a standalone binary. In that scenario, having `kube-downscaler` as the default value in `ExcludedNamespaces` still makes sense and i think it is necessary.

Looking at the chart I think we have 2 choices:

The first one is the one you already implemented, this would leave 

```yaml
 {{- with .Values.excludedNamespaces }}
 EXCLUDE_NAMESPACES: '{{ join "," . }}'
 {{- end }}
```

The second one would override the code default `kube-downscaler` with the `.Release.Namespace`. So even if the user doesn't install the chart inside a namespace called `kube-downscaler`, and EXCLUDED_NAMESPACES is empty the installation would be completely fine.

It is possible to use the `coalesce` operator that says: "when you have multiple values, return the first non-empty one", so in the case depicted below if  `.Values.excludedNamespaces` is empty, the default would be `list .Release.Namespace "kube-system"` (evaluated as `EXCLUDE_NAMESPACES: .Release.Namespace,kube-system`

```yaml
EXCLUDE_NAMESPACES: '{{ join "," (coalesce .Values.excludedNamespaces (list .Release.Namespace "kube-system")) }}'
```

samuel-esp:
Hey @gaetanars, I think for the moment we can proceed with solution 2

website/content/docs/3 - helm-chart/2 - values/16 - excludedNamespaces.mdx
```diff
@@ -39,4 +39,11 @@ Will result in all workloads in the namespaces `kube-downscaler`, `kube-system`
 
 :::
 
+:::note
+
+Setting `excludedNamespaces: []` does not clear the exclusion list.
+The chart omits the `EXCLUDE_NAMESPACES` env var when the list is empty, so the binary falls back to its built-in default (`kube-system`, `kube-downscaler`). To truly exclude nothing, set a regex that cannot match any namespace name, such as `["$^"]`.
```
samuel-esp: Specify that when this field is empty, the default values are `kube-system `and `Release.namespace` for the chart. If the tool is used and installed outside the chart, when EXCLUDED_NAMESPACES is not present, the binary defaults to `kube-system` and `kube-downscaler`

deployments/chart/templates/configmap.yaml
```diff
@@ -4,7 +4,9 @@ metadata:
   name: {{ .Values.configMap.name }}
   namespace: {{ .Release.Namespace }}
 data:
-  EXCLUDE_NAMESPACES: '{{- join "," .Values.excludedNamespaces }}'
+  {{- with .Values.excludedNamespaces }}
+  EXCLUDE_NAMESPACES: '{{ join "," . }}'
+  {{- end }}
```
samuel-esp: modify to set `kube-system` and `Release.namespace` as the default values when `.Values.excludedNamespaces` is empty

samuel-esp (changes requested):
When the .Values.excludedNamespaces

gaetanars:
Done — switched to solution 2 in c32a3bb. The configmap now uses `coalesce` so an empty `excludedNamespaces` falls back to `[Release.Namespace, kube-system]`, and the MDX doc has been updated to reflect that the chart defaults differ from the standalone binary defaults. Thanks for the guidance @samuel-esp!

samuel-esp:
Hey @gaetanars can you run pre-commit to validate the continuous integration? A commit will trigger it, even an empty one. Once it's green I will merge it

samuel-esp (approved):
LGTM

PR #455, opened 2026-05-13
Feat/Add minimum-replicas value to skip downscaling small workloads

CasvanDongen:
## Motivation

We run a multi-tenant Kubernetes cluster used by a lot of development teams. Some of them temporarily scale their deployments down to 0 replicas for their own reasons, and we don't want the downscaler to mess with that.

We're running with `--downtime-replicas=1` right now, which means the downscaler sets every matched workload to 1 during downtime. So a deployment a team had manually scaled to 0 actually gets scaled back up to 1, which is exactly the wrong direction.

This PR adds a `minimum-replicas` threshold. If a workload has fewer current replicas then the threshold, the downscaler skips it. So with `--downtime-replicas=1 --minimum-replicas=1`:

- 0 replicas: left alone (used to get scaled up to 1, which was the actual bug for us)
- 1 replica: already at the target, nothing to do
- 2 replicas: scaled to 1

## Changes

- New `MinimumReplicas` field on `Scope`, with the usual scope-hierarchy resolver (`Scopes.GetMinimumReplicas()`).
- New `--minimum-replicas` flag and `downscaler/minimum-replicas` annotation, reusing the existing `ReplicasValue` parser.
- `Workload.ScaleDown` now takes a second `minimumReplicas` arg. Only `replicaScaledWorkload` actually uses it; daemonsets, PDBs and suspend-scaled workloads ignore it.
- When the workload is below the threshold, `ScaleDown` returns `(nil, nil)` and the caller counts it as excluded.
- Docs: added a `Minimum Replicas` entry in `Values.mdx` and a link from each of the three scope pages.

## Tests Done

- `go test -race ./...` and `golangci-lint run` are both green.
- Added unit tests for annotation parsing, scope-precedence resolution, the `ScaleDown` threshold logic (above/at/below), and one integration test through `scanWorkload`.

### DIFF
--- a/cmd/kubedownscaler/main.go
+++ b/cmd/kubedownscaler/main.go
@@ -414,6 +414,8 @@ func scaleWorkloads(
 }
 
 // scaleWorkload scales the given workload according to the given wanted scaling state.
+//
+//nolint:cyclop // one branch per Scaling state, splitting it up doesn't really help
 func scaleWorkload(
 	scaling values.Scaling,
 	workload scalable.Workload,
@@ -461,9 +463,18 @@ setting different scaling states at the same time (e.g. downtime-period and upti
 			return fmt.Errorf("failed to get downscale replicas: %w", err)
 		}
 
-		savedResources, err := client.DownscaleWorkload(downscaleReplicas, workload, ctx)
+		minimumReplicas := scopes.GetMinimumReplicas()
+
+		savedResources, err := client.DownscaleWorkload(downscaleReplicas, minimumReplicas, workload, ctx)
 		if err != nil {
 			return fmt.Errorf("failed to downscale workload: %w", err)
+		}
+
+		if savedResources == nil {
+			slog.Debug("workload is below minimum replicas, skipping", "workload", workload.GetName(), "namespace", workload.GetNamespace())
+			workloadNamespaceMetrics.IncrementExcludedWorkloadsCount()
+
+			return nil
 		}
 
 		workloadNamespaceMetrics.IncrementDownscaledWorkloadsCount()
--- a/cmd/kubedownscaler/main_test.go
+++ b/cmd/kubedownscaler/main_test.go
@@ -26,12 +26,17 @@ func (m *MockClient) GetNamespaceAnnotations(namespace string, ctx context.Conte
 }
 
 func (m *MockClient) DownscaleWorkload(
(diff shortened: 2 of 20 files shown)


samuel-esp:
Hey @CasvanDongen thanks for your contribution, I see the problem and thanks for pointing it out we'll fix that behavior!

I just gave a quick look at the code, I didn't go into deep details yet but I was wondering if there is a reason behind the introduction of the new scope value

For example in the replicaScaledWorkload function refactoring this condition from

```go
if currentReplicasInt32 == downscaleReplicasInt32{
// don't scale
...
}

```
to 

```go
if currentReplicasInt32 <= downscaleReplicasInt32 {
	var originalReplicasInt32 int32
	var isOriginalReplicasSet bool

	originalReplicasInt32, isOriginalReplicasSet, err = getOriginalReplicasInt32(r)
	if err != nil {
		return metrics.NewSavedResources(0, 0), err
	}

	if !isOriginalReplicasSet {
		if currentReplicasInt32 < downscaleReplicasInt32{
            slog.Debug("workload is below minimum replicas, skipping", "workload", r.GetName(), "namespace", r.GetNamespace())
		}else{
			slog.Debug("workload is already at target scale down replicas, skipping", "workload", r.GetName(), "namespace", r.GetNamespace())
		}
		return metrics.NewSavedResources(0, 0), nil
	}

	savedResources := r.getSavedResourcesRequests(originalReplicasInt32 - downscaleReplicasInt32)

	slog.Debug("workload is already scaled down, skipping", "workload", r.GetName(), "namespace", r.GetNamespace())

	return savedResources, nil
}
```

wouldn't have produced the same result as yours (avoiding that downscaler bounces back replicas from 0 to the value defined in downscale-replicas)? My fear is that having 2 similar arguments would produce confusion

CasvanDongen:
I was under the impression this was intended behaviour. After checking py-kube-downscaler project I can see it isn't, so this is really just a bug. I'll close this MR and open a focused bug fix instead. Apologies for the noise.

samuel-esp:
> I was under the impression this was intended behaviour. After checking py-kube-downscaler project I can see it isn't, so this is really just a bug. I'll close this MR and open a focused bug fix instead. Apologies for the noise.

Don't worry at all! Thanks a lot for noticing this, we'll wait for your bug fix PR then 😄

PR #456, opened 2026-05-13
Fix/Skip workloads below downtime-replicas

CasvanDongen:
## Motivation

We run a multi-tenant Kubernetes cluster used by a lot of development teams. Some of them temporarily scale their deployments down to 0 replicas for their own reasons, and we don't want the downscaler to mess with that.

We're running with --downtime-replicas=1 right now, which means the downscaler sets every matched workload to 1 during downtime. So a deployment a team had manually scaled to 0 actually gets scaled back up to 1, which is exactly the wrong direction.

I checked py-kube-downscaler and it doesn't have this issue. Its downscale branch is gated by `replicas > downtime_replicas`, so worklaods at or below the target are skipped. This PR fixes the bug which was initially worked around in #455.

## Changes

- Fixed the bug in `replicaScaledWorkload.ScaleDown`.
- Added a regression test.

## Tests Done

- `go test -race ./...` and `golangci-lint run` both green.


samuel-esp (approved):
LGTM

PR #445, opened 2026-04-17
Fix/Improve Update Operations

samuel-esp:
## Motivation

This PR resolves the unintended behaviors described in #444.

## Changes

- Update operations will now be performed only when scaling state needs an update
- Ownership tracking was corrected for `cronJobs`, `jobs`, `daemonsets`. Resources already in the desired scaling state will no longer be managed by the downscaler

## Tests Done

- Unit Tests

PR #460, opened 2026-05-20
perf(client): skip Update API call when workload is already in desired state

gaetanars:
## Problem

Each scan iteration currently issues a Kubernetes `Update` API call against every targeted workload — **even when the workload is already in the desired state** (already downscaled, or already at its original replicas).

Concrete consequences:

- **Audit log spam**: in a cluster with hundreds of managed workloads, every scan interval (default 30s) produces hundreds of audit entries that record no actual change.
- **resourceVersion churn**: every workload's `resourceVersion` is incremented on every scan, which can confuse other controllers and trip optimistic concurrency loops elsewhere.
- **Unnecessary API server load**: full `Update` requests with their admission/validation pipelines run for nothing.

The existing short-circuit inside `ScaleDown` ([replicaScaledWorkloads.go:83-102](https://github.com/gaetanars/GoKubeDownscaler/blob/main/internal/pkg/scalable/replicaScaledWorkloads.go#L83-L102)) only covers a narrow case: workload already at the target replica count **and** no `original-replicas` annotation. In the much more common case (already downscaled **with** annotation set, or any CronJob/Job already suspended, or any DaemonSet already with the match-none label), the in-memory mutation is a no-op but `client.Update()` is still called.

## Fix

Snapshot the workload via the existing `Workload.Copy()` method **before** mutating, then compare with `reflect.DeepEqual` **after**. If nothing actually changed, log a `debug` message and skip the `Update` API call entirely.

```go
original, copyErr := workload.Copy()
// ... ScaleDown / ScaleUp ...
if original != nil && reflect.DeepEqual(original, workload) {
    slog.Debug("workload already in desired state, skipping update", ...)
    return savedResources, nil
}
err = workload.Update(c.clientsets, ctx)
```

### Why this approach

| Approach considered | Verdict |
|---|---|
| Add `changed bool` return to `ScaleUp`/`ScaleDown` on the `Workload` interface | Rejected — touches 12 implementations + admission controller + ~28 tests |
| Switch `Update` to `Patch` (JSON merge / strategic) | Rejected — does **not** remove the audit log entry; root problem persists |
| **Snapshot + `reflect.DeepEqual` in the client (this PR)** | **Chosen** — zero interface change, automatic coverage for any future `Workload` implementation |

### Properties

- **Generic**: works for every resource type without special-casing (Deployment, StatefulSet, HPA, KEDA `ScaledObject`, CronJob, DaemonSet, PDB, Argo Rollout, Prometheus, AutoscalingRunnerSet, Zalando Stack, Job).
- **Safe fallback**: if `Copy()` ever returns an error, we fall back to the previous behavior (unconditional `Update`). No regression risk.
- **Cheap**: one deep-copy + one `DeepEqual` per workload per scan. Negligible compared to the network round-trip we're avoiding.

## Secondary change

`Clientsets.Kubernetes` widened from `*kubernetes.Clientset` to `kubernetes.Interface`. All call sites already use only interface methods, so this is transparent at runtime. It enables injecting `k8s.io/client-go/kubernetes/fake` in tests.

## Tests

New `internal/api/kubernetes/client_test.go` with four cases asserting the exact number of `update` actions recorded by the fake clientset:

| Scenario | Expected `Update` calls |
|---|---|
| Deployment already at 0 replicas with correct `original-replicas` annotation → `DownscaleWorkload(0)` | **0** |
| Deployment at 5 replicas → `DownscaleWorkload(0)` | 1 |
| Deployment at 5 replicas, no annotation → `UpscaleWorkload` | **0** |
| Deployment at 0 replicas with annotation → `UpscaleWorkload` | 1 |

## Validation

- `go test --cover ./...` — all packages green
- `go vet ./...` — clean

## Test plan

- [ ] CI green
- [ ] Manual test on a dev cluster: deploy with several already-downscaled workloads, observe that audit log entries on the second+ scan drop to zero for unchanged workloads
- [ ] Confirm `--dry-run` still logs `"would have sent update"` only for workloads that would actually change

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/api/kubernetes/client.go
+++ b/internal/api/kubernetes/client.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"log/slog"
 	"os"
+	"reflect"
 	"strings"
 	"sync"
 	"time"
@@ -242,9 +243,29 @@ func (c client) DownscaleWorkload(
 	workload scalable.Workload,
 	ctx context.Context,
 ) (*metrics.SavedResources, error) {
+	original, copyErr := workload.Copy()
+	if copyErr != nil {
+		slog.Debug(
+			"failed to snapshot workload before scale down, will update unconditionally",
+			"workload", workload.GetName(),
+			"namespace", workload.GetNamespace(),
+			"error", copyErr,
+		)
+	}
+
 	savedResources, err := workload.ScaleDown(replicas)
 	if err != nil {
 		return metrics.NewSavedResources(0, 0), fmt.Errorf("failed to set the workload into a scaled down state: %w", err)
+	}
+
+	if original != nil && reflect.DeepEqual(original, workload) {
+		slog.Debug(
+			"workload already in desired scaled down state, skipping update",
+			"workload", workload.GetName(),
+			"namespace", workload.GetNamespace(),
+		)
+
+		return savedResources, nil
 	}
 
 	if c.dryRun {
@@ -269,9 +290,29 @@ func (c client) DownscaleWorkload(
 
 // UpscaleWorkload upscales the workload to the original replicas.
 func (c client) UpscaleWorkload(workload scalable.Workload, ctx context.Context) error {
+	original, copyErr := workload.Copy()
+	if copyErr != nil {
+		slog.Debug(
+			"failed to snapshot workload before scale up, will update unconditionally",
+			"workload", workload.GetName(),
+			"namespace", workload.GetNamespace(),
+			"error", copyErr,
+		)
+	}
+
 	err := workload.ScaleUp()
 	if err != nil {
 		return fmt.Errorf("failed to set the workload into a scaled up state: %w", err)
+	}
+
+	if original != nil && reflect.DeepEqual(original, workload) {
+		slog.Debug(
+			"workload already in desired scaled up state, skipping update",
+			"workload", workload.GetName(),
+			"namespace", workload.GetNamespace(),
+		)
+
+		return nil
 	}
 
 	if c.dryRun {
--- a/internal/api/kubernetes/client_test.go
+++ b/internal/api/kubernetes/client_test.go
@@ -1,0 +1,116 @@
+package kubernetes
+
+import (
+	"context"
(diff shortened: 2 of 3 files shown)


samuel-esp:
@gaetanars, thanks for pointing it out, it was addressed in this PR that should be reviewed soon #445. It has our highest priority at the moment

gaetanars:
Thanks for the pointer to #445 — I read through it carefully. I think the two PRs overlap on the symptom (no-op `Update` calls) but actually solve different layers:

- **#445** changes the `Workload` contract so each implementation declares whether an update is needed. It also fixes the ownership-tracking bugs in `cronJobs` / `jobs` / `daemonsets`, which is out of scope here.
- **#460** keeps the contract untouched and relies on `Copy()` + `reflect.DeepEqual` at the client layer, so the skip is automatic for any current or future workload type.

Even after #445 lands, a `DeepEqual` short-circuit would still catch:

1. **Future `Workload` implementations** that forget to compute the `bool` correctly — the client-layer check acts as a safety net.
2. **Drift cases** where a `ScaleDown` / `ScaleUp` implementation mutates a sub-field (annotation, label) but the returned `bool` disagrees with the actual in-memory diff, or vice versa.

|  | #445 | #460 |
|---|---|---|
| Surface of change | Interface + 12 impls + admission + ~28 tests | 1 client file + tests |
| Coverage for future `Workload` types | Manual (each new impl must return the bool correctly) | Automatic |
| Also fixes ownership bugs (`cronJobs` / `jobs` / `daemonsets`) | Yes | No — orthogonal |
| Risk surface | Logic duplicated per impl | Single chokepoint, safe fallback if `Copy()` errors |

Would you be open to landing **#445 first** (it's the deeper fix), and then keeping #460 as a thin client-side guard? Once the bool plumbing from #445 exists, this PR would shrink to ~30 LOC and serve purely as defense-in-depth. Happy to rebase on top of #445 and slim it down accordingly — or close this PR if you'd rather not carry the extra layer.

PR #469, opened 2026-06-02
New Resource: zalando postgresql (`acid.zalan.do/v1`)

### LINKED ISSUE #468: New Resource: zalando postgresql (`acid.zalan.do/v1`)
## Requested Resource

Support for Zalando postgres-operator `postgresql` custom resource (`acid.zalan.do/v1`).

Today there is no operator-native way to downscale a Zalando-managed Postgres with the downscaler. The only option is to include statefulsets, so the scanner patches `spec.replicas` to 0 on the operator-created sts. But `spec.numberOfInstances` on the postgresql CR - not the sts's replicas - is the operator's authoritative desired state, so the operator's sync loop detects the patched sts as drift (compareStatefulSetWith flags the replica mismatch) and re-applies its desired replica count on every reconcile. The scanner then patches it back to 0 on its next interval, and the two keep overwriting each other for the whole downtime window.

Scaling the Postgresql CR's own `spec.numberOfInstances` instead is operator-native: the operator treats "0 instances" as the desired state, reconciles the StatefulSet down itself, and stays quiet during downtime.

## Implementation Proposal

Add a postgresqls scalable that sets `spec.numberOfInstances` on the postgresql CR (a value int32). The resource is replica-scaled, so it slots into the existing replicaScaledWorkload wrapper like deployments/statefulsets/stacks.

### Further Details

Use case / benefit: operator-native, zero-log-noise downscaling of Zalando Postgres clusters during off-hours, without relying on the operator's StatefulSet-drift backoff.

Validation: I have a working implementation and validated it end-to-end against a live Zalando postgres-operator dev cluster:
- Scanner path - downscale patches `numberOfInstances` to 0 → operator scales the StatefulSet to 0, pods terminate; no could not connect warnings during the downtime window (vs. the constant warnings when scaling the STS directly). Upscale restores the original instance count from the downscaler/original-replicas annotation.
- The operator accepts numberOfInstances: 0 with no enforced minimum on the default config.
- Webhook path - the mutating webhook intercepts a CREATE/UPDATE of a postgresql and rewrites `numberOfInstances` to the downscale target inline, with no conflict against the operator's own webhooks.

Note on the field type: `numberOfInstances` is a value int32 (not a pointer), so unlike statefulsets/stacks there is no "replicas unset" case - the field always has a concrete value.

Note on enabling statefulsets + postgresqls together: postgresqls is the recommended resource for Zalando clusters. If both are enabled the downscaler also sees the operator's sts (the operator doesn't set ownerReferences by default), but this is harmless - the CR is authoritative and the operator reconciles the sts to match numberOfInstances regardless.

Transparency: the implementation was drafted with AI assistance and then reviewed and validated end-to-end against the live dev cluster above.

### Other Links/References

- Zalando postgres-operator: https://github.com/zalando/postgres-operator
- enable_owner_references (default false; enabling it disables the operator's delete-protection): https://github.com/zalando/postgres-operator/blob/master/docs/reference/operator_parameters.md
- Related ownerReference history: zalando/postgres-operator#498, zalando/postgres-operator#793
- Closest precedent in this repo: internal/pkg/scalable/autoscalingrunnersets.go (controller-runtime client + API-types-only import)

Art3mK:
## Motivation

Adds support for scaling Zalando postgres-operator `postgresql` resources (`acid.zalan.do/v1`), as discussed in #468.

Today there is no operator-native way to downscale a Zalando-managed Postgres with the downscaler. The only option is to include statefulsets, so the scanner patches `spec.replicas` to 0 on the operator-created sts. But `spec.numberOfInstances` on the postgresql CR - not the sts's replicas - is the operator's authoritative desired state, so the operator's sync loop detects the patched sts as drift (compareStatefulSetWith flags the replica mismatch) and re-applies its desired replica count on every reconcile. The scanner then patches it back to 0 on its next interval, and the two keep overwriting each other for the whole downtime window.

This change scales the postgresql CR's own `spec.numberOfInstances` instead - operator-native, so the operator treats "0 instances" as the desired state, reconciles the StatefulSet down itself, and stays quiet during downtime.

While validating the webhook path end-to-end on a real cluster, I found the chart's admission-controller (webhook) deployment was entirely non-functional under tight RBAC - the webhook pod couldn't start its cert rotation or watch its own resources. This PR also fixes those pre-existing chart bugs (see Changes), since they had to be fixed to test the webhook path for this feature.

## Changes

Feature - zalando postgresql support:
- New postgresqls scalable (internal/pkg/scalable/postgresqls.go): scales the Postgresql CR via `spec.numberOfInstances`, wrapped in the existing replicaScaledWorkload. Accessed through the controller-runtime client, importing only the postgres-operator API types package (github.com/zalando/postgres-operator/pkg/apis/acid.zalan.do/v1) - not its generated clientset - to avoid a k8s.io/client-go version conflict. Mirrors the autoscalingrunnersets scalable pattern.
- Registration (internal/pkg/scalable/workload.go): adds postgresqls to the resource map and postgresql to the admission-parse map.
- Scheme (internal/api/kubernetes/client.go): registers acidv1.AddToScheme on the controller-runtime scheme.
- Helm chart (deployments/chart/templates/_helpers.tpl, values.yaml): RBAC + webhook rules for postgresqls, plus the commented includedResources example.
- Dependency (go.mod, go.sum): adds github.com/zalando/postgres-operator (API types only; k8s.io/client-go unchanged).
- Tests (internal/pkg/scalable/postgresqls_test.go): set/get replicas (incl. the value-int32 no-unset case), saved-resource calculation (nil/partial/unparseable requests), Copy, Compare, scale-down/up round-trip, parsePostgresqlFromBytes.
- Docs (website/content/docs/2 - downscaler/3 - Workload Types.mdx, homepage SupportedResources.tsx).

Webhook chart fixes (pre-existing bugs found while testing the webhook path; each is its own commit):
- fix(chart): use webhook image tag for webhook deployment - the webhook Deployment built its image tag from .Values.image.tag (the main image) instead of .Values.webhookController.image.tag, which was defined but never consumed. Broke any setup pinning the webhook to a distinct tag/registry.
- fix(chart): allow webhook SA to create its TLS secret - the webhook SA's create on secrets was scoped with resourceNames; Kubernetes ignores resourceNames for create, so cert rotation got secrets is forbidden. Split into an unrestricted create + the name-scoped read/modify verbs.
- fix(chart): allow webhook SA to list/watch mutatingwebhookconfigurations - the reflector must list the collection before watching, but list was missing and the rule was resourceNames-scoped (which can't apply to list). Added an unrestricted list/watch rule.
- fix(chart): allow webhook SA to list/watch secrets - same resourceNames/list issue for the secret reflector. Moved list/watch to the unrestricted rule.

## Tests Done

- go test ./... - all green, including the new postgresqls unit tests.
- golangci-lint run - clean (0 issues).
- helm lint + helm template - chart renders correctly with postgresqls in includedResources and with the webhook enabled (RBAC, webhook rules, and both deployment images render as expected).
- End-to-end against a live zalando postgres-operator dev cluster:
  - Scanner path: downscale patched numberOfInstances → 0, the operator scaled the StatefulSet to 0 and terminated pods, with no could not connect log noise during the downtime window. Upscale restored the original instance count from the downscaler/original-replicas annotation.
  - Confirmed the operator accepts numberOfInstances: 0 (no enforced minimum on the default config).
  - Webhook path: after the chart RBAC fixes above, the webhook pod started cleanly (cert rotation + CA injection working); a CREATE/UPDATE of a Postgresql was intercepted and numberOfInstances rewritten to the downscale target inline, with no conflict against the operator's own webhooks.

### DIFF
--- a/deployments/chart/templates/_helpers.tpl
+++ b/deployments/chart/templates/_helpers.tpl
@@ -143,7 +143,6 @@ Create defined permissions for the webhook clusterrole
     - webhook.kube-downscaler.k8s
   verbs:
     - get
-    - watch
     - patch
     - update
 {{- end }}
--- a/deployments/chart/templates/_helpers.tpl
+++ b/deployments/chart/templates/_helpers.tpl
@@ -80,15 +80,18 @@ Create selector label for the webhook
 Create defined permissions for the webhook role
 */}}
 {{- define "go-kube-downscaler.webhookController.namespace.permissions" -}}
-# "create" cannot be restricted by resourceNames (the object does not exist yet, so the
-# API server ignores resourceNames-scoped rules for create). It must be a separate
-# unrestricted rule, scoped here only by the namespace of this Role.
+# "create", "list" and "watch" cannot be restricted by resourceNames: create has no object
+# to name-match yet, and the controller's reflector lists+watches the collection before it
+# can filter by name. These must be unrestricted, scoped only by this Role's namespace. The
+# name-scoped rule keeps the verbs that act on our specific secret (get/update/patch).
 - apiGroups:
     - ""
   resources:
     - secrets
   verbs:
     - create
+    - list
+    - watch
 - apiGroups:
     - ""
   resources:
@@ -96,11 +99,9 @@ Create defined permissions for the webhook role
   resourceNames:
     - {{ include "go-kube-downscaler.fullname" . }}-webhook
   verbs:
+    - get
     - update
     - patch
-    - get
-    - watch
-    - list
 - apiGroups:
     - coordination.k8s.io
   resources:
--- a/deployments/chart/templates/_helpers.tpl
+++ b/deployments/chart/templates/_helpers.tpl
@@ -80,6 +80,15 @@ Create selector label for the webhook
 Create defined permissions for the webhook role
 */}}
 {{- define "go-kube-downscaler.webhookController.namespace.permissions" -}}
+# "create" cannot be restricted by resourceNames (the object does not exist yet, so the
+# API server ignores resourceNames-scoped rules for create). It must be a separate
+# unrestricted rule, scoped here only by the namespace of this Role.
+- apiGroups:
+    - ""
+  resources:
+    - secrets
+  verbs:
+    - create
 - apiGroups:
     - ""
   resources:
@@ -87,7 +96,6 @@ Create defined permissions for the webhook role
   resourceNames:
     - {{ include "go-kube-downscaler.fullname" . }}-webhook
   verbs:
-    - create
     - update
     - patch
     - get
--- a/deployments/chart/templates/_helpers.tpl
+++ b/deployments/chart/templates/_helpers.tpl
@@ -285,6 +285,16 @@ Create defined permissions for roles
     - list
     - update
 {{- end }}
+{{- if eq $resource "postgresqls" }}
+- apiGroups:
+    - acid.zalan.do
+  resources:
+    - postgresqls
+  verbs:
+    - get
+    - list
+    - update
+{{- end }}
 {{- end }}
 {{- end }}
 
@@ -424,6 +434,17 @@ Create webhook resources
     - "UPDATE"
   resources:
     - autoscalingrunnersets
+{{ end -}}
+{{ if eq $resource "postgresqls" -}}
+- apiGroups:
+    - acid.zalan.do
+  apiVersions:
+    - "*"
+  operations:
+    - "CREATE"
+    - "UPDATE"
+  resources:
+    - postgresqls
 {{ end -}}
 {{ end -}}
 {{- end }}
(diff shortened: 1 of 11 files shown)


samuel-esp:
Hey @Art3mK thank you for your discoveries and for your contribution. Been busy weeks, I'll review it as soon as possible!

Art3mK:
@samuel-esp gentle reminder 😁

deployments/chart/templates/_helpers.tpl
samuel-esp: Thanks for pointing out that list/watch operations cannot be restricted by resourceNames, I didn't know this but for list it actually makes sense. At this point I think this "watch" permission here could be deleted since it would be redundant

samuel-esp (changes requested):
@Art3mK Thanks for your PR, overall it looks good to me. Just a small permission to delete because it looks like it is redudant after adding lines 128-137

deployments/chart/templates/_helpers.tpl
Art3mK: removed

samuel-esp (approved):
LGTM

PR #476, opened 2026-06-23
fix: don't skip ScaledObjects with undefined current replicas

### LINKED ISSUE #475: Bug: ScaledObjects not downscaled in 1.3.4
<details>

- Version: 1.3.4
- Installed via: Helm

</details>

## Description

After upgrading 1.3.3 → 1.3.4, KEDA ScaledObjects stop being downscaled. Confirmed on a cluster running 1.3.4 with `--downtime-replicas=1` and an active downtime window: 0 of 199 ScaledObjects were paused by the downscaler.

Cause: #456 (`14c173b`) changed the scale-down guard in `replicaScaledWorkload.ScaleDown` from `==` to `<=`:

https://github.com/caas-team/GoKubeDownscaler/blob/main/internal/pkg/scalable/replicaScaledWorkloads.go#L85

```go
if currentReplicasInt32 <= downscaleReplicasInt32 {
    // treats workload as already at/below target, skips scaling
}
```

This guard runs for every replica-scaled workload and collides with the "unset" sentinel ScaledObjects use.

### ScaledObjects (always broken)

A ScaledObject reports current replicas from the `autoscaling.keda.sh/paused-replicas` annotation. When the annotation is absent — i.e. any ScaledObject not already paused — `getReplicas()` returns the undefined sentinel `util.Undefined` = `-1` (`scaledobjects.go#L68-L82`, `util/util.go#L4`). Since `-1 <=` any non-negative `downtime-replicas`, every not-yet-paused ScaledObject matches the guard and is skipped with `workload is at or below target scale down replicas, skipping`, so it is never paused. This is independent of the configured `downtime-replicas`.

This is the supported path for downscaling KEDA workloads (the downscaler pauses the ScaledObject, not the KEDA-managed HPA), so KEDA workloads no longer scale down at all.

## Expected Behavior

ScaledObjects are downscaled during downtime, as in 1.3.3, and workloads previously scaled down remain scaled down. 

## Steps To Reproduce

1. Install 1.3.4.
2. Create a ScaledObject without the `paused-replicas` annotation.
3. Enter a downtime window.
4. It is not paused/downscaled; debug log shows `workload is at or below target scale down replicas, skipping`.

## Proposal

Add a regression test for a ScaledObject with no `paused-replicas` annotation.

### Other Links/References

- Introduced by #456 (`14c173b`)

dakotaharden:
## Description

Fixes #475.

#456 changed the scale-down guard in `replicaScaledWorkload.ScaleDown` from `==` to `<=`, so a workload already **below** the downtime target is skipped:

```go
if currentReplicasInt32 <= downscaleReplicasInt32 {
    // treat as already at/below target, skip
}
```

This also matches the `util.Undefined` (`-1`) sentinel that a KEDA `ScaledObject` reports from `getReplicas()` when it has no `autoscaling.keda.sh/paused-replicas` annotation (`scaledobjects.go`). Since `-1 <=` any non-negative `downtime-replicas`, **every not-yet-paused ScaledObject is skipped and never downscaled** — confirmed on a 1.3.4 cluster where 0 of 199 ScaledObjects were paused during an active downtime window.

## Change

Exclude the undefined sentinel from the skip check:

```go
if currentReplicasInt32 != util.Undefined && currentReplicasInt32 <= downscaleReplicasInt32 {
```

A workload whose current replicas are undefined (`-1`) now proceeds to be scaled down, the sentinel is recorded as `downscaler/original-replicas`, and scale-up restores it (deleting the paused-replicas annotation) via the existing `util.Undefined` handling in `setReplicas`/`ReplicasValue.Set`.

Only `scaledObject.getReplicas()` can return `util.Undefined`; every other workload type returns a real replica count or errors, so their behavior is unchanged. The intent of #456 (skip workloads genuinely below the downtime target) is preserved.

## Testing

Added `TestReplicaScaledWorkload_ScaleDown_ScaledObjectUndefinedReplicas`, covering the full scale-down → scale-up round-trip for a ScaledObject with no paused-replicas annotation. `go test ./internal/pkg/scalable/` passes.

### DIFF
--- a/internal/pkg/scalable/replicaScaledWorkloads_test.go
+++ b/internal/pkg/scalable/replicaScaledWorkloads_test.go
@@ -4,7 +4,9 @@ import (
 	"errors"
 	"testing"
 
+	"github.com/caas-team/gokubedownscaler/internal/pkg/util"
 	"github.com/caas-team/gokubedownscaler/internal/pkg/values"
+	kedav1alpha1 "github.com/kedacore/keda/v2/apis/keda/v1alpha1"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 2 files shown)


internal/pkg/scalable/replicaScaledWorkloads_test.go
```diff
@@ -226,3 +228,50 @@ func TestReplicaScaledWorkload_ScaleDown(t *testing.T) {
...
+	require.NoError(t, err)
+	restoredInt32, err := restored.AsInt32()
+	require.NoError(t, err)
+	assert.Equal(t, int32(util.Undefined), restoredInt32,
+		"scale up must restore the undefined state (no paused-replicas annotation)")
+
+	_, hasOriginal := getOriginalReplicas(workload)
+	var unsetErr *OriginalReplicasUnsetError
+	assert.ErrorAs(t, hasOriginal, &unsetErr, "original-replicas annotation must be removed after scale up")
+}
```
samuel-esp, line 302: Everything here is fine, just a small thing, could you refactor this to a table driven test? It would be easier to read. Look at the `TestReplicaScaledWorkload_ScaleUp` and `TestReplicaScaledWorkload_ScaleDown` tests above for reference

samuel-esp (changes requested):
Hey @dakotaharden, thanks a lot for your contribution and for finding this bug. The PR is fine, could you just refactor the test to be a "table driven" test? If you can't just request me the review again and I'll approve it, I'll refactor this later on

samuel-esp (approved):
LGTM

PR #483, opened 2026-07-02
Feat/Ignore Stale Exclude-Until Annotations When Setting Scopes

### LINKED ISSUE #482: Avoid Setting Stale Exclude-Until Annotations To Scopes
### Problem to Solve

When the downscaler encounters a downscaler/exclude-until annotation whose timestamp is in the past, it still sets its value in the scope.

Let's take this example:

ConfigMap is set like this:

```yaml
apiVersion: v1
data:
  DEFAULT_UPTIME: never
```

Namespace is set with an exclude-annotation (in the future) like this:

```yaml
apiVersion: v1
kind: Namespace
metadata:
  annotations:
    downscaler/exclude-until: "2026-08-31T23:00:00Z"
```

Workload is set with an exclude-annotation (in the past) like this:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  annotations:
    downscaler/exclude-until: "2024-12-25T22:30:00Z"
```

In this case, the Downscaler will downscale the workload because the exclusion evaluates to `false`. This can lead to unexpected behavior: although the namespace is technically excluded and should remain up and running, the workload is still downscaled because it has an higher-priority exclusion which is evaluated as `false` even though the annotation is technically stale. 

This is something is not happening with Py-Kube-Downscaler (as far as I remember)

### Proposal

The option may be either of these, or both:

1. If an exclude-until annotation is set to a time in the past, it should be ignored when setting the scope value.
2. The downscaler should check and remove stale annotations

For the moment I would start with option 1

### Who Can Address the Issue

- Go devs

samuel-esp:
## Motivation

As explained in #482 when the downscaler encounters a `downscaler/exclude-until` annotation whose timestamp is in the past, it still sets its value in the scope. This could lead to potential unwanted behaviors

## Changes

- Refactored `GetExcluded` function in `scopes.go`; `downscaler/exclude-until` annotations won't be considered anymore if the annotations is in the past
- Added unit tests dedicated to this case

## Tests Done

- Unit Tests

### DIFF
--- a/internal/pkg/values/scope.go
+++ b/internal/pkg/values/scope.go
@@ -325,7 +325,7 @@ func (s Scopes) GetExcluded(scopes Scopes) bool {
 		}
 
 		if !scope.ExcludeUntil.After(time.Now()) {
-			// ExcludeUntil is expired, continue checking the next scope
+			// ExcludeUntil is expired, continue checking the next scop
 			continue
 		}
 
--- a/internal/pkg/values/scope.go
+++ b/internal/pkg/values/scope.go
@@ -324,11 +324,12 @@ func (s Scopes) GetExcluded(scopes Scopes) bool {
 			continue
 		}
 
-		if scope.ExcludeUntil.After(time.Now()) {
-			return true
+		if !scope.ExcludeUntil.After(time.Now()) {
+			// ExcludeUntil is expired, continue checking the next scope
+			continue
 		}
 
-		// ExcludeUntil is expired, continue checking the next scope
+		return true
 	}
 
 	return false
--- a/internal/pkg/values/scope.go
+++ b/internal/pkg/values/scope.go
@@ -328,7 +328,7 @@ func (s Scopes) GetExcluded(scopes Scopes) bool {
 			return true
 		}
 
-		break
+		// ExcludeUntil is expired, continue checking the next scope
 	}
 
 	return false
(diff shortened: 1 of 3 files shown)


internal/pkg/values/scope.go
```diff
@@ -320,7 +320,7 @@ func (s Scopes) GetExcluded(scopes Scopes) bool {
 			return true
 		}
 
-		break
+		// ExcludeUntil is expired, continue checking the next scope
```
jonathan-mayer:
```suggestion
		if !scope.ExcludeUntil.After(time.Now()) {
			// ExcludeUntil is expired, continue checking the next scope
			continue
		}

		return true
```

just to keep the guard clause readability

internal/pkg/values/scopeParser.go
```diff
@@ -204,7 +204,9 @@ func (s *Scope) GetScopeFromAnnotations( //nolint: funlen,gocognit,cyclop,gocycl
 			return err
 		}
 
-		s.ExcludeUntil = &excludeUntil
+		if excludeUntil.After(time.Now()) {
+			s.ExcludeUntil = &excludeUntil
+		}
```
jonathan-mayer: this is kind of redundant and could cause confusion in future changes

jonathan-mayer (changes requested):

internal/pkg/values/scopeParser.go
jonathan-mayer: @samuel-esp was it intentional resolving this without changing it?

samuel-esp: No sorry, I missed that. I'll send over the commit fixing that after launch

samuel-esp: solved by bringing back the original statement

jonathan-mayer (approved):

